PR #3, opened 2019-01-08
PT-163053747 Add missing type check of init function

UlfNorell:
- [PT-163053747](https://www.pivotaltracker.com/story/show/163053747)


hanssv:
There needs to be a bump PR in ~~`epoch`~~ `aeternity` repo?

UlfNorell:
@hanssv make that part of your PR? Since it's related.

PR #17, opened 2019-01-25
PT-163478903 builtin bits type

UlfNorell:
Replaces bit arithmetic operations with a builtin `bits` type.


UlfNorell:
Tests are in https://github.com/aeternity/aeternity/pull/2077

PR #25, opened 2019-02-08
PT-156466783 namespaces

UlfNorell:
Pivotal: [PT-156466783](https://www.pivotaltracker.com/story/show/156466783)
Protocol: https://github.com/aeternity/protocol/pull/306

You can now do things like
```
namespace MyLib =
  function myfun(x) = x + 1

contract MyContract =
  function foo(x) = MyLib.myfun(x - 1)
```

This also includes a massive overhaul of the type checker, which fixes among other things type declarations not being checked properly.


nikita-fuchs:
Whenever adding/changing something regarding types, could @rvirding please be notified so it can be considered for the ACI generator? Cool feature !

PR #27, opened 2019-02-20
Skeleton for Fate backend

happi:


rebar.config
```diff
@@ -1,8 +1,11 @@
 {erl_opts, [debug_info]}.
 
 {deps, [ {aebytecode, {git, "https://github.com/aeternity/aebytecode.git",
-                            {branch,"fortuna"}}}
+                             {branch,"fortuna"}}}
        , {getopt, "1.0.1"}
+        % waiting for https://github.com/jlouis/enacl/pull/40 to be merged
```
hanssv: This should not be here?

happi: Removed enacl.

rebar.config
```diff
@@ -1,14 +1,14 @@
 {erl_opts, [debug_info]}.
 
 {deps, [ {aebytecode, {git, "https://github.com/aeternity/aebytecode.git",
-                            {branch,"fortuna"}}}
+                             {tag,"v2.0.1"}}}
```
hanssv: This reminds me again, we need to think about versioning (and documentation, and release notes?!)...

rebar.config
```diff
@@ -25,3 +25,11 @@
             {plt_apps, all_deps},
             {base_plt_apps, [erts, kernel, stdlib, crypto, mnesia]}
            ]}.
+
+{relx, [{release, {aesophia, "2.0.0"},
```
hanssv: What are we releasing?

PR #28, opened 2019-02-25
Add function to decode VM values to Sophia abstract syntax

UlfNorell:
To be used by some swanky compiler HTTP endpoint.


rvirding:
Could we get an interface function which allows us to not recompile the contract every time?

UlfNorell:
> Could we get an interface function which allows us to not recompile the contract every time?

What would be the point of that? Keeping the source code around is a lot easier for the user than keeping all the type information stored and tracking if it's up to date.

Also it's not compiling, only type checking.

PR #30, opened 2019-02-27
To sophia value revisited

hanssv:
Making the calldata creation and call result decoding easier to use.

### DIFF
--- a/src/aeso_compiler.erl
+++ b/src/aeso_compiler.erl
@@ -11,8 +11,8 @@
 -export([ file/1
         , file/2
         , from_string/2
-        , check_call/2
-        , create_calldata/3
+        , check_call/4
+        , create_calldata/4
         , version/0
         , sophia_type_to_typerep/1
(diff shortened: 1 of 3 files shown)


src/aeso_compiler.erl
```diff
@@ -145,40 +146,71 @@ check_call(ContractString, Options) ->
...
         end
     catch
         error:{parse_errors, Errors} ->
             {error, join_errors("Parse errors", Errors, fun (E) -> E end)};
         error:{type_errors, Errors} ->
             {error, join_errors("Type errors", Errors, fun (E) -> E end)};
-        error:{badmatch, {error, missing_decode_function}} ->
-            {error, join_errors("Type errors", ["missing __decode function"],
+        error:{badmatch, {error, missing_function}} ->
+            {error, join_errors("Type errors", ["no function: '" ++ FunName ++ "'"],
```
ThomasArts: Beautiful error message :-).

PR #31, opened 2019-02-27
New aci generator

rvirding:
Add an ACI generator to the compiler.

### DIFF
--- a/src/aeso_aci.erl
+++ b/src/aeso_aci.erl
@@ -136,8 +136,9 @@ decode_contract(#{<<"name">> := Name,
      %% decode_tdefs(Ts),
      decode_funcs(Fs)].
 
-decode_funcs(Fs) -> [ decode_func(F) || F <- Fs].
+decode_funcs(Fs) -> [ decode_func(F) || F <- Fs ].
 
+decode_func(#{<<"name">> := <<"init">>}) -> [];
 decode_func(#{<<"name">> := Name,<<"arguments">> := As,<<"type">> := T}) ->
     ["  function"," ",io_lib:format("~s", [Name])," : ",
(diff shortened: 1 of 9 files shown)


src/aeso_aci.erl
```diff
@@ -0,0 +1,259 @@
...
+    %% Try and return something sensible here!
+    case aeso_parser:string(Text) of
+        %% Yay, it worked!
+        {ok, Contract} -> Contract;
+        %% Scan errors.
+        {error, {Pos, scan_error}} ->
+            parse_error(Pos, "scan error");
+        {error, {Pos, scan_error_no_state}} ->
+            parse_error(Pos, "scan error");
+        %% Parse errors.
```
hanssv: The parse errors looks out of date...

src/aeso_aci.erl
```diff
@@ -0,0 +1,276 @@
...
+encode(ContractString, Options) ->
+    try
+        Ast = parse(ContractString, Options),
+        %%io:format("~p\n", [Ast]),
+        %% aeso_ast:pp(Ast),
+        TypedAst = aeso_ast_infer_types:infer(Ast, Options),
+        %% io:format("~p\n", [TypedAst]),
+        %% aeso_ast:pp_typed(TypedAst),
+        %% We find and look at the last contract.
+        Contract = lists:last(TypedAst),
```
UlfNorell: This won't work if you are using types from other namespaces/contracts. For instance
```
contract OtherContract =
  datatype some_type = This | That
  function stuff : some_type => int

contract MyContract =
  function do_the_stuff(r : OtherContract, x : OtherContract.some_type) =
    r.stuff(x)
```
but I guess that's not really in scope for this PR.

PR #32, opened 2019-02-28
Fix handling of init in create_calldata

hanssv:


src/aeso_compiler.erl
```diff
@@ -107,12 +107,27 @@ join_errors(Prefix, Errors, Pfun) ->
...
+%%       a special return type (typerep, T)
 -spec check_call(string(), string(), [string()], options()) -> {ok, string(), {[Type], Type | any}, [term()]} | {error, term()}
     when Type :: term().
-check_call(ContractString0, FunName, Args, Options) ->
+check_call(Source, "init" = FunName, Args, Options) ->
+    PatchFun = fun(T) -> {tuple, [typerep, T]} end,
+    case check_call(Source, FunName, Args, Options, PatchFun) of
+        {error, _} when Args == [] ->
+            %% Try with default init-function
+            check_call(insert_init_function(Source, Options), FunName, Args, Options, PatchFun);
```
UlfNorell: Will this not cause strange errors when you have an init function, but some other error?

A nicer approach might be to parse the contract and look for the init function, and only insert a default init if you don't find it.

hanssv: Yet another approach would be to return the first error when the second `check_call` also fail 
:thinking:

PR #46, opened 2019-03-15
Remove fragile file-based versioning

lucafavatella:


VERSION
```diff
@@ -1 +0,0 @@
-2.0.0
```
lucafavatella: Why is the VERSION file required in this repo?

hanssv: The VERSION was needed in two places originally, but that has been fixed since so just removing it makes sense 👍

PR #52, opened 2019-04-03
PT-163022973 Make a new aci generator

rvirding:
Add the new ACI generator which also converts type to JSON.

[PT-163022973](https://www.pivotaltracker.com/story/show/163022973)


src/aeso_aci.erl
```diff
@@ -59,7 +73,7 @@ encode(ContractString, Options) ->
                     T <- sort_decls(contract_types(Contract)) ],
         Fdefs = [ encode_func(F) || F <- sort_decls(contract_funcs(Contract)),
                                     not is_private_func(F) ],
-        Jmap = [{<<"contract">>, [{<<"name">>, list_to_binary(Cname)},
+        Jmap = [{<<"contract">>, [{<<"name">>, encode_name(Cname)},
```
hanssv: I think `jsx` accepts atoms as strings, not just binaries... It makes the code easier to read?!

rvirding:
Yes, it will be used by `aesophia_http` ASAP so they can test it.

PR #55, opened 2019-04-11
Update commit hash aebytecode

ThomasArts:
New version of aebytecode available


thepiwo:
wouldn't it be better to work with tags?

ThomasArts:
> wouldn't it be better to work with tags?

From a security perspective it is better to work with commit hash. One can let a tag point to a different hash without core-dev team being informed. At once one would be running different code.

PR #58, opened 2019-04-19
Fixed letrec and letfun env management in infer_block

radrow:
The env was not updated while processing a block. This was causing following correct code not being accepted by the typechecker:
```
contract Test =
  function f() =
    let x() = 3
    x() 
```
Error: `Unbound variable x`.


radrow:
This do not completely fix the issue, it is only typechecker part. After that there are some errors in `aeso_ast_to_icode.erl` file that I couldn't handle

UlfNorell:
Superseded by #65.

PR #60, opened 2019-04-23
Revert bytes(N) from icode/vm-types

hanssv:
For backwards compatibility (mainly in primops) it isn't a great idea to introduce a new VM-type. Instead, encode `bytes(N)` as either `word` or `{tuple, [word,...]}`.

### DIFF
--- a/src/aeso_ast_to_icode.erl
+++ b/src/aeso_ast_to_icode.erl
@@ -696,11 +696,8 @@ ast_typerep({qid, _, Name}, Icode) ->
     lookup_type_id(Name, [], Icode);
 ast_typerep({con, _, _}, _) ->
     word;   %% Contract type
-ast_typerep({bytes_t, _, Len}, _) when Len =< 32 ->
-    word;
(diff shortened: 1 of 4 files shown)


src/aeso_ast_to_icode.erl
```diff
@@ -696,8 +696,11 @@ ast_typerep({qid, _, Name}, Icode) ->
     lookup_type_id(Name, [], Icode);
 ast_typerep({con, _, _}, _) ->
     word;   %% Contract type
+ast_typerep({bytes_t, _, Len}, _) when Len =< 32 ->
+    word;
 ast_typerep({bytes_t, _, Len}, _) ->
-    {bytes, Len};
+    Words = (31 + Len) div 32,
+    {tuple, lists:duplicate(Words, word)};
```
UlfNorell: Can be simplified using `bytes_t`:
```erlang
ast_typerep({bytes_t, _, Len}, _) -> bytes_t(Len);
```

PR #65, opened 2019-05-10
Letrec and letfun

UlfNorell:
Removes `let rec` from the language, and fixes type checking and compilation of `let fun`. Supersedes #58.


src/aeso_ast_infer_types.erl
```diff
@@ -840,7 +840,7 @@ check_special_funs(_, _) -> ok.
 
 typesig_to_fun_t({type_sig, Ann, Named, Args, Res}) -> {fun_t, Ann, Named, Args, Res}.
 
-infer_letrec(Env, {letrec, Attrs, Defs}) ->
+infer_letrec(Env, Defs) ->
```
radrow: Why isn't `infer_letrec` completely removed? What's its use?

radrow: Why isn't `infer_letrec` completely removed? What's its use?

UlfNorell: It's used to check the top-level functions in a contract.

hanssv:
Is this worth a note in the changelog?

PR #69, opened 2019-05-14
Check stateful annotations

UlfNorell:
- `stateful` [PT-162578475](https://www.pivotaltracker.com/story/show/162578475)
- `init` [PT-160750930](https://www.pivotaltracker.com/story/show/160750930)

cc https://github.com/aeternity/protocol/pull/352

Functions must be annotated as `stateful` in order to
- Update the contract state (using `put`)
- Call `Chain.spend` or other primitive functions that cost tokens
- Call an Oracle or AENS function that requires a signature
- Make a remote call with a non-zero value
- Construct a lambda calling a stateful function. We get away with checking statefulness at the construction site rather than the call site since lambdas cannot outlive a contract call (no passing/receiving lambdas to/from other contracts, no storing lambdas in the state).

It does not need to be stateful to
- Read the contract state
- Call another contract with value=0, even when the remote function is stateful

There is a separate check for `init` that it neither reads nor writes the state. It can be stateful in the other senses though.


src/aeso_ast_infer_types.erl
```diff
@@ -862,7 +867,9 @@ infer_letrec(Env, Defs) ->
     [print_typesig(S) || S <- TypeSigs],
     {TypeSigs, NewDefs}.
 
-infer_letfun(Env, {letfun, Attrib, Fun = {id, NameAttrib, Name}, Args, What, Body}) ->
+infer_letfun(Env0, {letfun, Attrib, Fun = {id, NameAttrib, Name}, Args, What, Body}) ->
+    Env = Env0#env{ stateful = aeso_syntax:get_ann(stateful, Attrib, false),
```
ThomasArts: does this really work? Attribute must then contain {stateful, true}, I guess.

UlfNorell:
```erlang
1> proplists:get_value(foo, [foo]).                                                                       
true
```

PR #72, opened 2019-05-27
Bad record compiler crash

UlfNorell:
Fail with a parse error instead of crashing the type checker on bad record expressions with nested fields like
```
  {x.y = 17}
```


happi:
Why was  test/contracts/withdrawal.aes  deleted?

UlfNorell:
> Why was test/contracts/withdrawal.aes deleted?

It's an old example written in fantasy Sophia when all we had was a parser.

PR #85, opened 2019-06-11
Fate crypto ops

hanssv:
Also add `contract_to_address` and make fate compiler backend also return a `map`.

### DIFF
--- a/src/aeso_fcode_to_fate.erl
+++ b/src/aeso_fcode_to_fate.erl
@@ -95,6 +95,11 @@
      Op =:= 'BITS_OR'         orelse
      Op =:= 'BITS_AND'        orelse
      Op =:= 'BITS_DIFF'       orelse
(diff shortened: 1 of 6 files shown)


src/aeso_fcode_to_fate.erl
```diff
@@ -95,6 +95,11 @@
      Op =:= 'BITS_OR'         orelse
      Op =:= 'BITS_AND'        orelse
      Op =:= 'BITS_DIFF'       orelse
+     Op =:= 'SHA3'            orelse
+     Op =:= 'SHA256'          orelse
+     Op =:= 'BLAKE2B'         orelse
+     Op =:= 'ECVERIFY'        orelse
+     Op =:= 'ECVERIFY_SECP256K1' orelse
```
UlfNorell: and `CONTRACT_TO_ADDRESS`

PR #87, opened 2019-06-12
Pt 166636909 avoid calldata collision

ThomasArts:


src/aeso_compiler.erl
```diff
@@ -211,13 +211,13 @@ check_call1(ContractString0, FunName, Args, Options) ->
...
-                %% case aeb_fate_code:symbol_identifier(<<"__call">>) of
-                ContractString = insert_call_function(ContractString0, FunName, Args, Options),
+                %% collect all hashes and compute the first name without hash collision to
+                SymbolHashes = maps:keys(aeb_fate_code:symbols(FateCode)),
+                CallName = first_none_match(?CALL_NAME, SymbolHashes,
+                                            lists:seq($1, $9) ++ lists:seq($A, $Z) ++ lists:seq($a, $z)),
+                ContractString = insert_call_function(ContractString0, CallName, FunName, Args, Options),
                 #{fcode := Fcode} = string_to_fcode(ContractString, Options),
-                #{args := CallArgs} = maps:get({entrypoint, <<"__call">>}, maps:get(functions, Fcode)),
+                #{args := CallArgs} = maps:get({entrypoint, list_to_binary(CallName)}, maps:get(functions, Fcode)),
```
hanssv: This will only work for nulllary functions (since the added function will always have 0 arguments...) This is a problem in the original PR not introduced here.

test/aeso_calldata_tests.erl
```diff
@@ -0,0 +1,75 @@
+%%% -*- erlang-indent-level:4; indent-tabs-mode: nil -*-
```
hanssv: We should have a test that actually creates calldata for something with arguments :-)

ThomasArts: Yes, done that now

src/aeso_fcode_to_fate.erl
```diff
@@ -231,7 +231,28 @@ lookup_var(#env{vars = Vars}, X) ->
...
+        {account_pubkey, K}  -> aeb_fate_data:make_address(K);
+        {contract_pubkey, K} -> aeb_fate_data:make_contract(K);
+        {oracle_pubkey, K}   -> aeb_fate_data:make_oracle(K);
+        {oracle_query_id, _} -> ?TODO(fate_oracle_query_id_value)
+     end;
+term_to_fate(nil) ->
+    aeb_fate_data:make_list([]);
+term_to_fate({con, _Ar, I, As}) ->
+    %% what is Ar
+    aeb_fate_data:make_list([term_to_fate(I) | [ term_to_fate(A) || A<-As]]);
```
UlfNorell: `con` is not list cons, but variant type construction (`Ar` is the list of arities of the constructors, `I` is the tag index, and `As` is the constructor arguments). Cons is `{op, '::', [Hd, Tl]}`.

src/aeso_fcode_to_fate.erl
```diff
@@ -231,7 +231,28 @@ lookup_var(#env{vars = Vars}, X) ->
...
+term_to_fate({tuple, As}) ->
+    aeb_fate_data:make_tuple(list_to_tuple([ term_to_fate(A) || A<-As])).
```
UlfNorell: TODO:
- variant types
- negative integers (`{op, '-', [{lit, {int, N}}]}`)
- Bits (not working for aevm though)
- hash and signature literals (coordinate with @hanssv)

src/aeso_fcode_to_fate.erl
```diff
@@ -231,7 +231,28 @@ lookup_var(#env{vars = Vars}, X) ->
...
+
+
+
 to_scode(_Env, {lit, L}) ->
+    %% use term_to_fate here
```
UlfNorell: Or better: break out `lit_to_fate` from `term_to_fate` and use that.

test/aeso_calldata_tests.erl
hanssv: Where?

UlfNorell: I don't see the tests. You mean done, but not pushed?

src/aeso_fcode_to_fate.erl
ThomasArts: Thanks. Will check with @hanssv. Pushed the test for dutch_auction, counter and bitcoin_auth that take different arguments. Bitcoin_auth is not compilable yet, but when it is, test is ready.

UlfNorell: The test would look nicer if you defined a single contract with functions whose arguments cover all the cases in `term_to_fate`. Choosing a different existing test contract and defining a new unit test for each case is a bit clumsy, and it's hard to get an overview of the test coverage.

src/aeso_fcode_to_fate.erl
```diff
@@ -231,7 +231,30 @@ lookup_var(#env{vars = Vars}, X) ->
...
+        {contract_pubkey, K} -> aeb_fate_data:make_contract(K);
+        {oracle_pubkey, K}   -> aeb_fate_data:make_oracle(K);
+        {oracle_query_id, _} -> ?TODO(fate_oracle_query_id_value)
+     end;
+%% negative literals are parsed as 0 - N
+term_to_fate({op, '-', [{lit, {int, 0}}, {lit, {int, N}}]}) ->
+    aeb_fate_data:make_integer(-N);
+term_to_fate(nil) ->
+    aeb_fate_data:make_list([]);
+term_to_fate({op, '::', [Hd | Tl]}) ->
```
UlfNorell:
```suggestion
term_to_fate({op, '::', [Hd, Tl]}) ->
```

src/aeso_fcode_to_fate.erl
```diff
@@ -231,7 +231,30 @@ lookup_var(#env{vars = Vars}, X) ->
...
+    aeb_fate_data:make_list([term_to_fate(Hd) | [ term_to_fate(E) || E<-Tl]]);
```
UlfNorell:
```suggestion
    aeb_fate_data:make_list([term_to_fate(Hd) | term_to_fate(Tl)]);
```

UlfNorell: Works, since `make_list` is identity. Should really do `?FATE_LIST_VALUE` on the tail.

src/aeso_fcode_to_fate.erl
```diff
@@ -240,14 +240,17 @@ term_to_fate({lit, L}) ->
...
+    lit_to_fate(L);
 %% negative literals are parsed as 0 - N
 term_to_fate({op, '-', [{lit, {int, 0}}, {lit, {int, N}}]}) ->
     aeb_fate_data:make_integer(-N);
 term_to_fate(nil) ->
     aeb_fate_data:make_list([]);
-term_to_fate({op, '::', [Hd | Tl]}) ->
-    aeb_fate_data:make_list([term_to_fate(Hd) | [ term_to_fate(E) || E<-Tl]]);
+term_to_fate({op, '::', [Hd, Tl]}) ->
+    aeb_fate_data:make_list([term_to_fate(Hd) | aeb_fate_data:make_list(term_to_fate(Tl))]);
```
UlfNorell: The second `make_list` makes no sense. `term_to_fate` already adds a `make_list` so if anything that `make_list` should be stripped off (but `make_list` is currently a no-op, so...). I guess the most correct thing to do without messing with aebytecode (and adding `aeb_fate_data:X_value` functions) is to include `aeb_fate_data.hrl` and use ?FATE_LIST_VALUE.

ThomasArts: In that case I probably would just only use `term_to_fate(Tl)` because including that abstraction breaker in this code seems weird.

hanssv:
Signatures and hash are missing, but it is probably easier if I take care of them in a follow up PR.

src/aeso_fcode_to_fate.erl
UlfNorell: Remove the outer `make_list` as well then.

src/aeso_fcode_to_fate.erl
```diff
@@ -231,16 +231,36 @@ lookup_var(#env{vars = Vars}, X) ->
...
+term_to_fate(nil) ->
+    aeb_fate_data:make_list([]);
+term_to_fate({op, '::', [Hd, Tl]}) ->
+    %% The Tl will translate into a list, because FATE lists are just lists
+    [term_to_fate(Hd) | term_to_fate(Tl)];
+term_to_fate({tuple, As}) ->
+    aeb_fate_data:make_tuple(list_to_tuple([ term_to_fate(A) || A<-As]));
+term_to_fate({con, Ar, I, As}) ->
+    FateAs = [ term_to_fate(A) || A <- As ],
+    aeb_fate_data:make_variant(Ar, I, list_to_tuple(FateAs)).
```
UlfNorell: also missing: maps

PR #95, opened 2019-06-25
PT-166233670 FATE events

UlfNorell:
~~TODO:~~
- [x] merge https://github.com/aeternity/aebytecode/pull/58
- [x] handle bytes in AEVM events
- [x] more interesting types in test (oracles, bits, bytes, ...)

### DIFF
--- a/src/aeso_ast_to_fcode.erl
+++ b/src/aeso_ast_to_fcode.erl
@@ -130,7 +130,7 @@
 -type env() :: #{ type_env   := type_env(),
                   fun_env    := fun_env(),
                   con_env    := con_env(),
-                  event_type := aeso_syntax:typedef(),
+                  event_type => aeso_syntax:typedef(),
                   builtins   := builtins(),
                   options    := [option()],
                   context    => context(),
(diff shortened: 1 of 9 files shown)


src/aeso_ast_to_fcode.erl
```diff
@@ -783,6 +793,8 @@ op_builtins() ->
 
 builtin_to_fcode(require, [Cond, Msg]) ->
     make_if(Cond, {tuple, []}, {builtin, abort, [Msg]});
+builtin_to_fcode(chain_event, [Event]) ->
+    {def, event, [Event]};
```
hanssv: Any risk of name clash? I remember we talked about it 🤔

test/aeso_compiler_tests.erl
```diff
@@ -119,11 +119,7 @@ compilable_contracts() ->
      "bytes_to_x"
     ].
 
-not_yet_compilable(fate) ->
-    ["events",              %% events
-     "address_literals",    %% oracle_query_id literals
-     "address_chain"        %% Oracle.check_query
-    ];
+not_yet_compilable(fate) -> [];
```
hanssv: Yay 🍌 💃

src/aeso_ast_to_fcode.erl
UlfNorell: `event` is a special name here, user defined functions are `{entrypoint, Name}` or `{local_fun, Name}`. It turns into `Chain.event` in the fate code.

PR #101, opened 2019-06-28
Prepare version 3.2.0

hanssv:


CHANGELOG.md
```diff
@@ -9,6 +9,38 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+    // Exported
+    entrypoint exported_fun(x) = local_fun(x)
+    // Not exported
+    function local_fun(x) = x
+  ```
+  Functions in namespaces still use `function` (and `private function` for
+  private functions).
+- The return type of `Chain.block_hash(height)` has changed, it used to
+  be `int`, where `0` denoted an incorrect height. New return type is
+  `option(hash)`, where `none` represents an incorrect height.
```
happi: Is it not 'None'?

hanssv: Bah... Good point.

PR #105, opened 2019-07-09
Added list comprehensions and standard List, Option, Func, Pair, and Triple library

radrow:
## New syntax

Example test file
```haskell
contract ListComp =

    entrypoint sample1() = [1,2,3]
    entrypoint sample2() = [4,5]

    entrypoint l1() = [x | x <- sample1()]
    entrypoint l1_true() = [1,2,3]

    entrypoint l2() = [x + y | x <- sample1(), y <- sample2()]
    entrypoint l2_true() = [5,6,6,7,7,8]

    entrypoint l3() = [x ++ y | x <- [[":)"] | x <- [1,2]]
                              , y <- [[":("]]]
    entrypoint l3_true() = [[":)", ":("], [":)", ":("]]

    entrypoint l4() = [(a, b, c) | let is_pit(a, b, c) = a*a + b*b == c*c
                                 , let base = [1,2,3,4,5,6,7,8,9,10]
                                 , a <- base
                                 , b <- base, if (b >= a)
                                 , c <- base, if (c >= b)
                                 , if (is_pit(a, b, c))
                                 ]
    entrypoint l4_true() = [(3, 4, 5), (6, 8, 10)]

```

## Standard library

Standard libraries are included automatically. They take priority over user defined files and any attempt to redefine them will lead to an error.

### Developing
They are written as inlined strings in `aeso_stdlib.erl`, however in the future they should (IMO) be kept as separate `aes` files and at best compiled during aesophia build time. 

If standard library wants to import another one it just needs to be mentioned in `dependencies/1` function. The keyword `include` should not be used in stdlib code.

To preserve compatibility with existing tests all functions in stdlib should have explicitly defined types.

## Other changes

`expand_includes` in `aeso_parse` has yet another argument – `Included` which is a set of already included files that are going to be skipped. They are encoded as pairs of their filenames and hashes of their contents. This solution solves most problems with cyclic and transitive imports and allows to safely mix them up.
This argument is also included in `parse` in `aeso_compiler`.

I have added one more compiler option – `no_implicit_stdlib` which stops standard library being included automatically. This helps a bit with includes inside of the stdlib and fixes compatibility issues of already existing tests that didn't expect additional includes and is very handy in tests that just don't require stdlib which has really bad impact on overall compilation performance (it increases build time by about 1 second of a single contract on my PC).

## TODO

 - Move Sophia code out of the `erl` files++
 - Optimizations: for instance `[f(x) | x <- l]` could be replaced by `Lists.map`


UlfNorell:
> - Lack of List.aes include will cause compiler crash missing flat_map function – proper message should be written

A perhaps nicer alternative is to import it implicitly.

> - Compile standard library during build time

We have no machinery for separate compilation and I don't see that there would be any tangible benefits from it.

src/aeso_stdlib.erl
```diff
@@ -0,0 +1,199 @@
...
+
+  function const(x) = (y) => x
+
+  function comp(f, g) = (x) => f(g(x))
+
+  function iter(n, f) = iter_(n, f, (x) => x)
+  private function iter_(n, f, acc) =
+    if(n == 0) acc
+    elif(n == 1) comp(f, acc)
+    else iter_(n / 2, f, if(n mod 2 == 0) comp(acc, acc) else comp(f, comp(acc, acc)))
```
UlfNorell: This isn't right. `iter(2 ^ n, f) == f`. You want the recursive call to be
```suggestion
    else iter_(n / 2, comp(f, f), if(n mod 2 == 0) acc else comp(f, acc))
```
However I don't think this is really a function that we should put in the standard libarary.

src/aeso_stdlib.erl
```diff
@@ -0,0 +1,199 @@
...
+
+  function untuplify2(f) = (x, y) => f((x, y))
+  function untuplify3(f) = (x, y, z) => f((x, y, z))
+".
+
+std_list() ->
+"
+namespace List =
+
+  function empty(l) = switch(l)
```
UlfNorell: `is_empty`?

src/aeso_stdlib.erl
```diff
@@ -0,0 +1,199 @@
...
+    [] => true
+    _  => false
+
+
+  function first(l) = switch(l)
```
UlfNorell: `head`?

src/aeso_stdlib.erl
```diff
@@ -0,0 +1,199 @@
...
+    [x] => Some(x)
+    _::t => last(t)
+
+  function find(p, l) = switch(l)
+    [] => None
+    h::t => if(p(h)) Some(h) else find(p, t)
+
+  function nth(n, l) = switch(l)
+    h::t => if(n == 0) Some(h) else nth(n-1, t)
+    [] => None
```
UlfNorell: To make these `option`-valued functions convenient to use, we should should also have some standard functions on options (but let's leave that for another PR)

src/aeso_stdlib.erl
```diff
@@ -0,0 +1,199 @@
...
+  private function partition_(p, l, acc_t, acc_f) = switch(l)
+    [] => (reverse(acc_t), reverse(acc_f))
+    h::t => if(p(h)) partition_(p, t, h::acc_t, acc_f) else partition_(p, t, acc_t, h::acc_f)
+
+
+  function concats(ll) = foldr((l1, l2) => l1 ++ l2, [], ll)
+
+  function all(p, l) = foldl((prev, next) => prev && p(next), true, l)
+
+  function any(p, l) = foldl((prev, next) => prev || p(next), false, l)
```
UlfNorell: We probably want short-cut behaviour for these.

src/aeso_parser.erl
```diff
@@ -495,7 +501,7 @@ parse_pattern(E = {bool, _, _})   -> E;
 parse_pattern(E = {bytes, _, _})  -> E;
 parse_pattern(E = {string, _, _}) -> E;
 parse_pattern(E = {char, _, _})   -> E;
-parse_pattern(E) -> bad_expr_err("Not a valid pattern", E).
+parse_pattern(E) -> bad_expr_err("Not a valid patxstern", E).
```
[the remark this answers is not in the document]
radrow: emacs for the win :p fixed

src/aeso_ast_to_fcode.erl
```diff
@@ -354,7 +354,6 @@ make_let(Expr, Body) ->
             {'let', X, Expr, Body({var, X})}
     end.
 
--spec expr_to_fcode(env(), aeso_syntax:expr()) -> fexpr().
```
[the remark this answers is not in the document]
radrow: undeleted

src/aeso_stdlib.erl
radrow: What do you mean?

src/aeso_stdlib.erl
radrow: That's good idea. Maybe we should start an issue for it?

src/aeso_stdlib.erl
radrow: I know your point, but imo `first` will look more clear for a random user

src/aeso_stdlib.erl
radrow: fixed

src/aeso_stdlib.erl
radrow: My bad, sorry, fixed. I believe it can be used in some iterative algorithms and we don't gain much if we remove it

src/aeso_ast_infer_types.erl
```diff
@@ -2250,6 +2304,13 @@ pp_when({check_expr, Expr, Inferred0, Expected0}) ->
 pp_when({checking_init_type, Ann}) ->
     io_lib:format("when checking that 'init' returns a value of type 'state' at ~s\n",
                   [pp_loc(Ann)]);
+pp_when({list_comp, BindExpr, Inferred0, Expected0}) ->
```
radrow: Maybe I should consider cases of `comprehension_bind` and `comprehension_if`, @UlfNorell what do you think?

src/aeso_stdlib.erl
radrow: I have added some support

radrow:
> > * Lack of List.aes include will cause compiler crash missing flat_map function – proper message should be written
> 
> A perhaps nicer alternative is to import it implicitly.
> 
> > * Compile standard library during build time
> 
> We have no machinery for separate compilation and I don't see that there would be any tangible benefits from it.

The tests take radically more time. On the other hand I have provided option to skip stdlib – we may use it in tests that don't need it

src/aeso_stdlib.erl
UlfNorell: That `all` only evaluates `p` until it gets a `false`, in the same way that `false && x` doesn't evaluate `x`.

radrow: fixed

PR #107, opened 2019-07-19
Changed 'require' function return type to Unit

radrow:
`require` function could be used to exploit type system, because it was returning `'a` type instead of `()`:

bug copyright (c) Martin Grigorov
```
contract Identity =
  type state = ()
  entrypoint main(x : int) = x
  
  entrypoint sum(a: int, b: int) = a + b
  
  entrypoint test_sum_correct() =
        let result = 5
        require(sum(2,3)== 5, "Result of sum is incorrect!")
```


radrow:
cc @UlfNorell @hanssv

PR #106, opened 2019-07-10
Fix split_pat

radrow:
Due to bad `fsplit_pat` construction the following contract
```
contract XD =

   entrypoint tuplify3() = (t) => switch(t)
     (x, y, z) => 3
```
used not to compile

PR #111, opened 2019-07-27
Fixed pretty printing and pattern split

radrow:
Pretty printing was broken in few places due to type errors – one may check it by executing `pp_fexpr` on bodies of functions right after processing. There were also some bugs in managing patterns (`fpat` was sometimes mistaken with `fsplit_pat`). I have added one test that used to crash previous version.

Beside that I undo #106 because the "fix" I provided there turned out to be treatment of symptoms. This PR should fix that issue in proper way now.

### DIFF
--- a/src/aeso_ast_to_fcode.erl
+++ b/src/aeso_ast_to_fcode.erl
@@ -1373,7 +1373,7 @@ pp_fexpr({builtin_u, B, N}) ->
 pp_fexpr({builtin, B, As}) ->
     pp_call(pp_text(B), As);
 pp_fexpr({remote_u, Ct, Fun, Ar}) ->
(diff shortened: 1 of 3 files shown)


src/aeso_ast_to_fcode.erl
```diff
@@ -681,7 +681,7 @@ split_pat({con, As, I, Pats}) ->
     Xs = [fresh_name() || _ <- Pats],
     {{con, As, I, Xs}, Pats};
 split_pat({tuple, Pats}) ->
-    Xs = [{var, fresh_name()} || _ <- Pats],
+    Xs = [fresh_name() || _ <- Pats],
```
lucafavatella: I see in the second element of `{tuple, _}` no need for `var` according to the type. Ok.

radrow: It was confusing a bit because in `fpat` this `var` was mandatory but in `fsplit_pat` no

PR #125, opened 2019-08-16
PT-162578406 Add payable modifier

hanssv:
Needs https://github.com/aeternity/aebytecode/pull/69


uwiger:
Still need the WIP label?

src/aeso_icode.erl
```diff
@@ -103,6 +106,10 @@ new_env() ->
 set_name(Name, Icode) ->
     maps:put(contract_name, Name, Icode).
 
+-spec set_payable(boolean(), icode()) -> icode().
+set_payable(Name, Icode) ->
```
uwiger: `Name` is a boolean?

src/aeso_ast_to_icode.erl
```diff
@@ -17,11 +17,15 @@
...
-               _ -> gen_error(last_declaration_must_be_contract)
+    {Payable, Name} =
+        case lists:last(TypedTree) of
+            {contract, Attrs, {con, _, Con}, _} ->
+                {proplists:get_value(payable, Attrs, false), Con};
+            _ ->
+                gen_error(last_declaration_must_be_contract)
            end,
-    NewIcode = aeso_icode:set_name(Name, aeso_icode:new(Options)),
+    NewIcode = aeso_icode:set_payable(Payable,
```
uwiger: So here, `Payable` is definitely a boolean, but in `aeso_icode:set_payable/2`, the first argument is called `Name` (though typed as a boolean), which at least looking at this context, seems confusing.

src/aeso_icode.erl
hanssv: Copy pasted, should be `Payable` of course.

hanssv:
Now updated with the correct `aebytecode` commit

PR #131, opened 2019-08-22
Update CHANGELOG.md (stdlib)

radrow:


hanssv:
I covered List comprehensions in the CHANGELOG already in #130, and maybe let's skip StdLib until @UlfNorell has worked his magic there?

radrow:
@hanssv What kind of magic are we speaking about?

UlfNorell:
We were discussing getting rid of the implicit import and making `flat_map` for desugared list comprehensions a builtin. Also possibly precomputing the result of parsing the standard library modules to make it a bit faster.

radrow:
Is there an issue for this?

PR #135, opened 2019-08-28
Remove find_all from stdlib

radrow:
It was just a duplicated `filter`


dincho:
Is this already released ?

radrow:
@dincho current form of it is not included in the latest aeternity tag if this is what you are asking for

PR #136, opened 2019-08-29
Updated some functions, renamed some stuff, added from_to IN STDLIB

radrow:


priv/stdlib/List.aes
```diff
@@ -37,14 +37,18 @@ namespace List =
...
     h::t => find_indices_(p, t, n+1, if(p(h)) n::acc else acc)
 
-  function nth(n : int, l : list('a)) : option('a) = switch(l)
-    []   => None
-    h::t => if(n == 0) Some(h) else nth(n-1, t)
+  function nth(n : int, l : list('a)) : option('a) =
+    if(n < 0) None
+    else switch(l)
+      []   => None
+      h::t => if(n == 0) Some(h) else nth(n-1, t)
```
UlfNorell: What's the motivation for this change? The semantics hasn't changed and the function is slower for legitimate use cases.

priv/stdlib/List.aes
```diff
@@ -37,14 +37,18 @@ namespace List =
...
 
   /* Unsafe version of `nth` */
-  function get(n : int, l : list('a)) : 'a = switch(l)
-    [] => abort("Out of index get")
-    h::t => if(n == 0) h else get(n-1, t)
+  function get(n : int, l : list('a)) : 'a =
+    if(n < 0) abort("Negative index get")
+    else switch(l)
+      [] => abort("Out of index get")
+      h::t => if(n == 0) h else get(n-1, t)
```
UlfNorell: Same here. It doesn't seem worth the extra check just to get a special error for negative indices.

priv/stdlib/List.aes
```diff
@@ -53,6 +57,15 @@ namespace List =
     _::t => length_(t, acc + 1)
 
 
+  function from_to(a : int, b : int) : list(int) = from_to_(a, b, [])
```
UlfNorell:
```suggestion
  function from_to(a : int, b : int) : list(int) = [a..b]
```

priv/stdlib/List.aes
radrow: More meaningful abort message for example. However, the check `n < 0` could be after list traversal

radrow: Rollbacked

priv/stdlib/List.aes
radrow: Moved `if` to the `abort`. It will keep meaningful error message and not affect performance

priv/stdlib/List.aes
```diff
@@ -53,6 +55,15 @@ namespace List =
     _::t => length_(t, acc + 1)
 
 
+  function from_to(a : int, b : int) : list(int) = [a..b]
+  private function from_to_(a, b, acc) =
+    if (a > b) acc else from_to_(a, b - 1, b :: acc)
```
UlfNorell: This is no longer used.

radrow: Fixd

PR #142, opened 2019-09-05
More structured errors - also in aeso_compiler

hanssv:


src/aeso_code_errors.erl
```diff
@@ -8,7 +8,7 @@
 %%%-------------------------------------------------------------------
 -module(aeso_code_errors).
 
--export([format/1]).
+-export([format/1, pos/1]).
```
UlfNorell: I don't think you need this.

PR #158, opened 2019-09-23
Don't confuse variables and store registers in fate asm generation

UlfNorell:


UlfNorell:
That's not a bad idea, and wouldn't even be consensus breaking I guess.

Issue #170, opened 2019-11-11
CLI: Standard library files not reachable in escript


hanssv:
```
~/Quviq/Aeternity/aesophia_cli [git:*master]: cat /tmp/foo.aes 
include "List.aes"
contract C =

  entrypoint foo(f : int => bool, l : list(int)) =
    List.map(f, l)
~/Quviq/Aeternity/aesophia_cli [git:*master]: ./aesophia_cli /tmp/foo.aes 
Parse error in '/tmp/foo.aes' at line 1, col 1:
Couldn't find include file 'List.aes'
```

There are actually two problems
 - [ ] The `aesophis/priv/stdlib/*` is not included in the escript `aesophia_cli`
 - [ ] You can't access the files in an escript as if they are in the file system

PR #171, opened 2019-11-11
Add stdlib include handling when inside an escript

hanssv:
Part of #170

### DIFF
--- a/src/aeso_parser.erl
+++ b/src/aeso_parser.erl
@@ -619,11 +619,28 @@ read_file(File, Opts) ->
             case maps:get(binary_to_list(File), Files, not_found) of
                 not_found -> {error, not_found};
                 Src       -> {ok, Src}
+            end;
+        escript ->
+            try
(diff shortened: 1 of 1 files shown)


src/aeso_parser.erl
```diff
@@ -619,11 +619,28 @@ read_file(File, Opts) ->
...
+                {error, not_found}
             end
     end.
 
 stdlib_options() ->
-    [{include, {file_system, [aeso_stdlib:stdlib_include_path()]}}].
+    StdLibDir = aeso_stdlib:stdlib_include_path(),
+    case filelib:is_dir(StdLibDir) of
+        true  -> [{include, {file_system, [StdLibDir]}}];
+        false -> [{include, escript}]
```
tolbrino: This should not default to using the escript option, but instead also check that the execution environment is in an escript. The default should be no additional options when no directory can be found and the parser is not executed in an escript.

PR #178, opened 2019-11-18
changed type of client ttl to int option

skkw:


skkw:
> I see there are no test case changes, maybe we should have a test that covers compiling this new function ;-)

That's correct. I'll make a PR for this.

PR #182, opened 2019-11-25
GH-181 Prepare 4.1.0-rc1

UlfNorell:
Ref #181


tolbrino:
So upon release the notes for any release candidates are consolidated into the single release entry?

UlfNorell:
It felt weird to have the changelog for 4.0.0 split across all the release candidates. I don't think they should be considered released versions of the compiler.

tolbrino:
I agree. It might make sense to have the full list of changes for the final release and keep the separate entries for the candidates.

PR #205, opened 2019-12-10
FATE backend optimisations

UlfNorell:
Notably
- Use more than one state register: state records use separate registers for each component.
- Unbox singleton records, i.e. `record { x : int }` is compiled to `int`.


src/aeso_fcode_to_fate.erl
```diff
@@ -179,42 +182,45 @@ lit_to_fate(L) ->
...
-term_to_fate({op, '::', [Hd, Tl]}) ->
+term_to_fate(Env, {op, '::', [Hd, Tl]}) ->
     %% The Tl will translate into a list, because FATE lists are just lists
-    [term_to_fate(Hd) | term_to_fate(Tl)];
-term_to_fate({tuple, As}) ->
-    aeb_fate_data:make_tuple(list_to_tuple([ term_to_fate(A) || A<-As]));
-term_to_fate({con, Ar, I, As}) ->
-    FateAs = [ term_to_fate(A) || A <- As ],
+    [term_to_fate(Env, Hd) | term_to_fate(Env, Tl)];
+term_to_fate(Env, {tuple, As}) ->
```
ThomasArts: Possibly I do not really get all type layers, but {tuple, [A]} in this case would not translate to just A as in scode, but rather keep the FATE tuple notation.

Most likely on purpose.

UlfNorell: No, you're right. That probably needs a case for singletons.

UlfNorell: I'll try to make a test case.

src/aeso_fcode_to_fate.erl
```diff
@@ -645,11 +652,11 @@ pp_op(loop) -> "LOOP";
...
     aeb_fate_pp:format_op(I, #{}).
 
-pp_arg(?i(I))      -> io_lib:format("~w", [I]);
-pp_arg({arg, N})   -> io_lib:format("arg~p", [N]);
-pp_arg({store, N}) -> io_lib:format("store~p", [N]);
-pp_arg({var, N})   -> io_lib:format("var~p", [N]);
-pp_arg(?a)         -> "a".
+pp_arg(?i(I))    -> io_lib:format("~w", [I]);
+pp_arg({arg, N}) -> io_lib:format("arg~p", [N]);
+pp_arg(?s(N))    -> io_lib:format("store~p", [-N]);
```
ThomasArts: I see, the stores are negative numbers and you want to print "store1" for `{store, -1}`

src/aeso_fcode_to_fate.erl
UlfNorell: No, it's right as it is. The singleton optimisation happens in `ast_to_fcode` and `term_to_fate` works on `fcode` so it's already happened at that point.

src/aeso_fcode_to_fate.erl
UlfNorell: No, that's just wrong. Look at the definition of `desugar_arg`:
```erlang
desugar_arg(?s(N)) -> {var, -N};
desugar_arg(A) -> A.
```

UlfNorell:
> so most likely there is no consensus issue...

The compiler is not under consensus (aesophia is not even a dependency of the node!), so there is guaranteed to be no consensus issues.

PR #206, opened 2019-12-11
GH-194 let patterns

### LINKED ISSUE #194: Data deconstruction in let
*From the Sophia discussion at the Sofia workshop.*

For instance,
```
  let x :: xs = ys
  let (x, y) = pt
  let Some(x) = Map.lookup(k, m)
```

UlfNorell:
Fixes #194 

~~Work-in-progress. Sits on top of #205.~~

Also adds pattern matching is list comprehension generators
```sophia
function somes(xs : list(option('a))) = [ x | Some(x) <- xs ]
```

Also also
- fix pattern matching on negative integers
- compile FATE values to immediates (e.g. `[1, 2, 3]` is compiled to `PUSH [1, 2, 3]` instead of a series of `CONS` instructions)

### DIFF
--- a/test/contracts/let_patterns.aes
+++ b/test/contracts/let_patterns.aes
@@ -1,0 +1,13 @@
+contract LetPatterns =
+
+  record r = {x : int, y : int, b : bool}
+
+  entrypoint test() = foo([1, 0], (2, 3), Some(4), {x = 5, y = 6, b = false})
+
+  entrypoint foo(xs : list(int), p : int * int, some : option(int), r : r) =
+    let x :: _  = xs
+    let (a, b)  = p
+    let Some(n) = some
+    let {x = i, y = j} = r
+    x + a + b + n + i + j
+
--- a/test/contracts/let_patterns.aes
+++ b/test/contracts/let_patterns.aes
@@ -11,3 +11,5 @@ contract LetPatterns =
     let {x = i, y = j} = r
     x + a + b + n + i + j
 
+  entrypoint lc(xs : list(option(int))) : list(int) =
+    [ x | Some(x) <- xs ]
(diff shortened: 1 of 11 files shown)


test/contracts/let_patterns.aes
```diff
@@ -0,0 +1,13 @@
+contract LetPatterns =
```
tolbrino: There doesn't seem to be a test case for the match support in list comprehensions.

UlfNorell: Good catch. I added it to my temp test file instead of the actual test case.

test/contracts/let_patterns.aes
```diff
@@ -0,0 +1,15 @@
...
+
+  record r = {x : int, y : int, b : bool}
+
+  entrypoint test() = foo([1, 0], (2, 3), Some(4), {x = 5, y = 6, b = false})
+
+  entrypoint foo(xs : list(int), p : int * int, some : option(int), r : r) =
+    let x :: _  = xs
+    let (a, b)  = p
+    let Some(n) = some
```
tolbrino: So this will crash if `some = None`?

test/contracts/let_patterns.aes
```diff
@@ -0,0 +1,15 @@
...
+    let {x = i, y = j} = r
+    x + a + b + n + i + j
+
+  entrypoint lc(xs : list(option(int))) : list(int) =
+    [ x | Some(x) <- xs ]
```
tolbrino: Here the pattern match acts as a filter?

test/contracts/let_patterns.aes
UlfNorell: Yes, this is equivalent to
```sophia
switch(some)
  Some(n) => ...
```

test/contracts/let_patterns.aes
UlfNorell: Indeed
```sophia
lc([None, Some(1), Some(2)]) == [1, 2]
```

Issue #196, opened 2019-12-09
Pattern matching in function heads


UlfNorell:
*From the Sophia discussion at the Sofia workshop.*

UlfNorell:
The current syntax (mixing left-hand sides and type signatures) is not well suited for this. We could imagine allowing separate type signatures and multiple clauses:
```
  entrypoint
    foo : (int, string) => string
    foo(0, s) = s
    foo(n, s) = foo(n - 1, String.concat(s, s))
```

radrow:
I love this

nikita-fuchs:
I sense a strong smell of erlang in here - wouldn't this defacto lever out all strict type checking and allow for fallback functions taking whatever data ?

UlfNorell:
What, no! Why would you think this would change the type system? The above is the same as
```sophia
entrypoint foo(a : int, s : string) =
  switch(a)
    0 => s
    n => foo(n - 1, String.concat(s, s)
```

PR #210, opened 2020-01-14
GH-196 pattern matching lhs

UlfNorell:
Fixes GH-196

Some examples:
```sophia
  function from_some(Some(x)) = x

  function
    length : list('a) => int
    length([])      = 0
    length(_ :: xs) = 1 + length(xs)

  function
    append([], ys)      = ys
    append(x :: xs, ys) = x :: append(xs, ys)
```


UlfNorell:
After ac9e9716cb43370eba106762ac3fc3442d3cd60a it compiles to exactly the same code. The type checker desugars it to a switch on the tuple of all the arguments, so there's a little work to be done by the backend to get rid of the extra matching-a-tuple-against-a-tuple.

PR #215, opened 2020-02-07
Optimize stdlib

radrow:
Tail recursion is not always as rewarding as in xkcd commics. Changed functions have about 2 times less gas usage


radrow:
I did run the stdlib test from `aeternity` and it passes.

PR #216, opened 2020-02-07
Expose some functionalities for REPL

radrow:
I have created and exposed some utility functions. I need to be able to pin into certain parts of the compiler for the purpose of my REPL


radrow:
Moving to lima: https://github.com/aeternity/aesophia/pull/218

PR #218, opened 2020-02-07
Prepared for REPL usage

radrow:
I have created and exposed some utility functions. I need to be able to pin into certain parts of the compiler for the purpose of my REPL.

Beside that I enchanted the way of running the parser – there is now a function `run_parser/{2,3}` which takes a parser `/0`, a string and optionally option list. Also, I added a `keep_included` option for the parser to report which files were included along with their hashes. This is useful when you want to keep track of which files are in context (I am using it in REPL to display warnings about duplicated includes for example)

### DIFF
--- a/src/aeso_compiler.erl
+++ b/src/aeso_compiler.erl
@@ -38,6 +38,7 @@
                 | pp_assembler
                 | pp_bytecode
                 | no_code
+		| keep_included
                 | {backend, aevm | fate}
                 | {include, {file_system, [string()]} |
                             {explicit_files, #{string() => binary()}}}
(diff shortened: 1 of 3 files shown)


src/aeso_parser.erl
```diff
@@ -596,8 +604,13 @@ expand_includes(AST, Included, Opts) ->
              || File <- lists:usort(auto_imports(AST)) ] ++ AST,
     expand_includes(AST1, Included, [], Opts).
 
-expand_includes([], _Included, Acc, _Opts) ->
-    {ok, lists:reverse(Acc)};
+expand_includes([], Included, Acc, Opts) ->
+    case lists:member(keep_included, Opts) of
```
hanssv: `Opts` have type `aeso_compiler:options()` and that does not include `keep_included`!?

radrow: Good catch, will add

radrow:
After I faced problems with pattern definitions I need to rework it a bit. Let's postpone it until I remove WIP label

src/aeso_compiler.erl
```diff
@@ -38,6 +38,7 @@
                 | pp_assembler
                 | pp_bytecode
                 | no_code
+		| keep_included
```
hanssv:
```suggestion
		  | keep_included
```

PR #219, opened 2020-02-07
Restrict relation operators to ints

radrow:
I see no reason to compare strings and functions using `<` operator, especially when it does nothing but return `false`.

I am working now with a custom datatype that represents some numbers and I accidentally used these operators several times which leaded to irritating bugs.


radrow:
Tests need to be updated after we agree for it

UlfNorell:
They don't always return false. Comparison is lexicographic on the underlying FATE representation.

PR #225, opened 2020-02-13
Instant unification error on arguments count mismatch

### LINKED ISSUE #224: "no arguments" displayed as empty string
```
AESO> function g(x) = x + 1
AESO> g() 
Error:
1:1:Cannot unify int
         and 
when checking the application at line 1, column 1 of
  g : (int) => int
to arguments
```

Speaking of the
```
1:1:Cannot unify int
         and 
```
part. This is highly dangerous for the aesthetics

radrow:
fixes #224


radrow:
Further upgrades would require more effort. This PR should be okay as it is therefore

Issue #176, opened 2019-11-18
More functions on strings


UlfNorell:
Some suggestions
```
  String.split(i : int, s : string) : string * string
  String.to_list(s : string) : list(char)
  String.from_list(cs : list(char)) : string
  String.at(s : string, i : int) : char
```
cc @thepiwo @radrow

radrow:
May be also useful:
```
Char.to_int : char -> int
Char.from_int : int -> option?(char) 
```

hanssv:
Maybe `String.contains`, etc.

maptuhec:
Also useful:

```
String.to_upper_case 
String.to_lower_case
```

maptuhec:
`String.to_bytes`?

UlfNorell:
> `String.to_bytes`

What type would that have?

maptuhec:
This was a suggestion from @hanssv for the following issue. The ecrecover returns. bytes, and if you want to compare it to a string it will not be possible for now(Bytes.to_str  returns uppercase String) if the string is not passed in uppercase. So probably the type should be bytes. Just an issue that we have run into today.

Best,
Martin

UlfNorell:
The problem is that `bytes` have fixed length and you don't know the length of the string at compile time.

There could be a `Bytes.to_raw_string : bytes(n) => string` (or some better name) though, that gives you the bytes as they are with out hex-encoding.

I'm confused as to why you would have your address as an unencoded string though. It would contain lots of unprintable characters.

tolbrino:
Something like `String.to_int(s : string) : int` would also be helpful.

tolbrino:
And `String.tokens(s : string, c : char) : list(string)` would split the string into tokens with the given character used as limiter.

radrow:
@tolbrino `String.to_int(s : string) : option(int)`, let's be safe

radrow:
I just hope that the strings won't become a bad replacement for ADTs

hanssv:
Maybe even `String.to_int(s : string, base : int) : option(int)`?

PR #236, opened 2020-02-24
Gh 176 more string functions

hanssv:
Part of #176 

Needs https://github.com/aeternity/aebytecode/pull/91


test/contracts/state_handling.aes
```diff
@@ -1,3 +1,4 @@
+include "String.aes"
```
radrow: Why is this include necessary?

hanssv: It uses `String.length`

Issue #203, opened 2019-12-09
Runtime validation of contracts


UlfNorell:
Checking at runtime that a given contract indeed implements the expected interface.

UlfNorell:
Note that such a check would give you no protection against malicious contracts, and would only be a(n expensive) sanity check ruling out silly mistakes.

nikita-fuchs:
I've always been a little sceptical towards these interface checking endeavours, because people tend to believe in what they see - contracts then can easily pretend to do something that they actually won't, to say the least. It's a great feature for a general quick check, but the docs would need to read big warning signs here.

thepiwo:
@UlfNorell could you give an example of what issues this would catch?

UlfNorell:
Technically, what would be possible (but moderately expensive) is to check that there is a contract at a given address with a given set of entrypoints of expected types. However, contracts can lie about their types and a crash due to a missing entrypoint looks the same as a crash inside that entrypoint, so the only protection would be against people giving you bad contracts by accident.

I think a better feature would be to let you recover from failing contract calls using some try/catch mechanism. That would let you protect yourself also from malicious contracts, and would be much cheaper.

nikita-fuchs:
This is a really useful thing. I favor the try-catch approach as a leach solution, although the first one would also have its benefits.

UlfNorell:
Can you elaborate on the benefits of the first suggestion?

UlfNorell:
The solution we settled on is adding *protected calls*. If
```sophia
contract Remote =
  entrypoint foo : int => int
```

Then for `r : Remote`
```sophia
r.foo(19) : int  // as before
r.foo(protected = true, 19) : option(int)
```
If the remote call fails for any reason (the contract doesn't exist/doesn't have `foo`, or `foo` has the wrong type, or crashes/runs out of gas, etc.) the protected call returns `None`.

PR #241, opened 2020-03-02
Add support for protected contract calls

UlfNorell:
- Issue: aeternity/aesophia#203
- aeternity/aebytecode#93
- aeternity/aeternity#3162
- aeternity/protocol#467

### DIFF
--- a/src/aeso_ast_infer_types.erl
+++ b/src/aeso_ast_infer_types.erl
@@ -1372,7 +1372,7 @@ infer_expr(Env, {typed, As, Body, Type}) ->
     Type1 = check_type(Env, Type),
     {typed, _, NewBody, NewType} = check_expr(Env, Body, Type1),
     {typed, As, NewBody, NewType};
-infer_expr(Env, {app, Ann, Fun, Args0}) ->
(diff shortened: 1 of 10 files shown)


src/aeso_ast_infer_types.erl
```diff
@@ -2185,6 +2250,9 @@ unify1(_Env, {qcon, _, Name}, {qcon, _, Name}, _When) ->
     true;
 unify1(_Env, {bytes_t, _, Len}, {bytes_t, _, Len}, _When) ->
     true;
+unify1(Env, {if_t, _, {id, _, Id}, Then1, Else1}, {if_t, _, {id, _, Id}, Then2, Else2}, When) ->
+    unify(Env, Then1, Then2, When) andalso
+    unify(Env, Else1, Else2, When);
```
radrow: Why doesn't `{if_t, {bool, _, true}, T, _}` unify with `T`?

UlfNorell: The argument to `if_t` is always an `id()`. As soon as we know the value the `if_t` goes away.

PR #250, opened 2020-03-23
Updated CHANGELOG

radrow:
I went through the diff and the git log, but if I missed something feel free to push


radrow:
"We update the changelog on every release, not on each commit..." ~ Hans

PR #256, opened 2020-04-02
Merge lima – 4.3.0

radrow:


hanssv:
There is a bit of documentation missing I think... The documentation has moved from protocol to aesophia. Just merging like this would leave the documentation in an inconsistent state.

radrow:
I considered writing a documentation a separate step from this merge. Or else I can merge it into yet another branch, fix documentation there and then put it on master. I am not sure which approach is the most SOTA

tolbrino:
> There is a bit of documentation missing I think... The documentation has moved from protocol to aesophia. Just merging like this would leave the documentation in an inconsistent state.

So the documentation changes from master would have to be merged to lima first, and then this step here can happen?

radrow:
No, the master contains undocumented changes that are not included in lima and therefore not in the lima's documentation. What is needed to be done, is to write the documentation for master on master.

hanssv:
Nothing is undocumented, but the documentation is written in the iris branch of the protocol-repo. It should be moved to aesophia, preferably in connection to this merge.

radrow:
The docs have been updated

radrow:
Okay, added AENS. Also, readded delegation signature minichapter

PR #261, opened 2020-05-15
Ban empty contracts, ban function blocks with mismatched declarations

### LINKED ISSUE #258: Function blocks accept clauses with different names
I don't think it should be legal. Also, makes my REPL unhappy.
```
contract C =
  entrypoint
    f() = 123
    g() = 123
```

### LINKED ISSUE #260: Compiler crashes on empty file
As I remember there is an attempt to handle this situation, but it seems to fail

radrow:
fixes #260 
fixes #258

### DIFF
--- a/src/aeso_ast_infer_types.erl
+++ b/src/aeso_ast_infer_types.erl
@@ -564,6 +564,10 @@ init_env(_Options) -> global_env().
 
 -spec infer(aeso_syntax:ast(), list(option())) ->
     aeso_syntax:ast() | {env(), aeso_syntax:ast()}.
+infer([], Options) ->
+    create_type_errors(),
+    type_error({no_decls, proplists:get_value(src_file, Options, no_file)}),
+    destroy_and_report_type_errors(init_env(Options));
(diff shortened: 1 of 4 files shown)


src/aeso_ast_infer_types.erl
```diff
@@ -626,7 +630,9 @@ infer_contract_top(Env, Kind, Defs0, _Options) ->
 %% a list of definitions.
 -spec infer_contract(env(), main_contract | contract | namespace, [aeso_syntax:decl()]) -> {env(), [aeso_syntax:decl()]}.
 infer_contract(Env0, What, Defs0) ->
+    create_type_errors(),
     Defs = process_blocks(Defs0),
+    destroy_and_report_type_errors(Env0),
```
radrow, line 635: @UlfNorell could you verify? I could not leave these type errors undestroyed here because it was causing the whole compiler to crash, had to create separate ones for this particular case.

UlfNorell: lgtm

PR #284, opened 2020-10-09
Debug mode to turn off hermetization

### LINKED ISSUE #277: Option to turn off hermetization
For the testing purposes it would be very valuable to be able to access hidden functions of contracts and namespaces

radrow:
Added `debug_mode` option which exposes all functions in all namespaces and main contract to ease the process of unit testing. Meaning of this flag can be later extended to some additional features.

fixes #277

### DIFF
--- a/src/aeso_aci.erl
+++ b/src/aeso_aci.erl
@@ -69,7 +69,7 @@ do_contract_interface(Type, ContractString, Options) ->
     try
         Ast = aeso_compiler:parse(ContractString, Options),
         %% io:format("~p\n", [Ast]),
-        {TypedAst, _} = aeso_ast_infer_types:infer(Ast, [dont_unfold]),
+        {TypedAst, _} = aeso_ast_infer_types:infer(Ast, [dont_unfold | Options]),
         %% io:format("~p\n", [TypedAst]),
         from_typed_ast(Type, TypedAst)
     catch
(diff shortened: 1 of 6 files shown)


hanssv:
Friday evening so not looking at the PR in detail. But this is really dangerous - imagine if you forget to switch off the option and deploy - is there a clever way to make it safer? 🤔

radrow:
I don't think it is a real issue - this flag is supposed to be used only by debuggers and sdks. If someone would be about to deploy a contract in debug mode then their interface should warn them about it

radrow:
We can also add a pragma for network id and require additional flag if someone is going to deploy debug contract to the mainnet. But this would require some assertion of the net id from inside of the bytecode to make it work reasonably...

hanssv:
I'm not sure it is worth it, I'm normally all for giving users/people good tools for shooting themselves in the foot - it's just that crypto people seem to be so talented in screwing up...

gorbak25:
I agree that exposing a debug mode gives users a new way to kill their app.
On the other hand I see 2 reasons why this is a good idea:
- Unit tests of internals - let's say I've implemented my own qsort routine in my contract and I want to test it without exporting it as an entrypoint.
- A bytecode validation infrastructure needs to be set up sooner or later - a contract compiled in debug mode won't validate

It's a security/usability trade-off.

radrow:
Anyway, nobody uses the compiler directly. There is always a frontend to it that is responsible for handling flags and configurations. If it is not implemented there, the flag won't be even usable for regular users.

hanssv (approved):

gorbak25 (approved):

Issue #234, opened 2020-02-20
"using namespace" qualification


radrow:
Depends on #233 

Extension of the above. This really helps finding "where the heck is (...) defined". Speaking of the possibility to include only specific parts of the namespace.
```
contract Con =
  using List hiding [flat_map, map]
  using Option for [flat_map, map2, map] 
```

Issue #233, opened 2020-02-20
"using namespace"


radrow:
When you are heavily using some namespace it becomes very annoying to mention its name every time. Also, one could make aliases for namespaces:
```
include "List.aes"
include "Option.aes"
contract Con =
  using List
  using Option as O
  entrypoint f(x) = O.concat_option(sum(x)) // `sum` originally in `List` 
```
The naming collisions would throw type errors and ask for qualification.

PR #338, opened 2021-08-23
Implement loading namespaces with the `using` keyword

ghallak:
Fixes: #233 
Fixes: #234

### DIFF
--- a/src/aeso_ast_infer_types.erl
+++ b/src/aeso_ast_infer_types.erl
@@ -340,19 +340,9 @@ visible_in_used_namespaces(UsedNamespaces, QName) ->
             IsVisible = fun(Namespace) ->
                             case Namespace of
                                 {_, _, {for, Names}} ->
-                                    case lists:member(Name, Names) of
-                                        true ->
-                                            true;
-                                        false ->
-                                            false
-                                    end;
+                                    lists:member(Name, Names);
                                 {_, _, {hiding, Names}} ->
-                                    case lists:member(Name, Names) of
-                                        true ->
-                                            false;
-                                        false ->
-                                            true
-                                    end;
+                                    not lists:member(Name, Names);
                                 _ ->
                                     true
                             end
--- a/src/aeso_ast_infer_types.erl
+++ b/src/aeso_ast_infer_types.erl
@@ -103,6 +103,8 @@
 -type typesig() :: {type_sig, aeso_syntax:ann(), type_constraints(), [aeso_syntax:named_arg_t()], [type()], type()}.
 
 -type namespace_alias() :: none | name().
+-type namespace_parts() :: none | {for, [name()]} | {hiding, [name()]}.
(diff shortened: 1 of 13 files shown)


radrow (approved):
Super!

src/aeso_ast_infer_types.erl
```diff
@@ -312,9 +317,48 @@ bind_contract({Contract, Ann, Id, Contents}, Env)
...
+        Namespaces ->
+            IsVisible = fun(Namespace) ->
+                            case Namespace of
+                                {_, _, {for, Names}} ->
+                                    case lists:member(Name, Names) of
+                                        true ->
+                                            true;
+                                        false ->
+                                            false
+                                    end;
```
radrow, line 348:
```suggestion
                                    lists:member(Name, Names)
```

src/aeso_ast_infer_types.erl
```diff
@@ -312,9 +317,48 @@ bind_contract({Contract, Ann, Id, Contents}, Env)
...
+                                {_, _, {hiding, Names}} ->
+                                    case lists:member(Name, Names) of
+                                        true ->
+                                            false;
+                                        false ->
+                                            true
+                                    end;
```
radrow, line 355:
```suggestion
                                    not lists:member(Name, Names)
```

src/aeso_ast_infer_types.erl
ghallak: I'm surprised that I wrote it this way. Will change.

radrow (approved):

hanssv:
Sorry I'm late to the party, but I was thinking about the example `using_namespace_ambiguous_name` - isn't the problem that you alias both `Xa` and `Xb` to `A` and the error is really on the line `using Xb as A` - the example reports two (other) errors? Otherwise it looks good to me.

Also there is a trivial conflict in documentation it seems.

ghallak:
> isn't the problem that you alias both `Xa` and `Xb` to `A` and the error is really on the line `using Xb as A` - the example reports two (other) errors?

@hanssv I would say that it's not an error that two namespaces have the same alias, as these namespaces might have functions with different names:

```
namespace Xa = function f() = 1
namespace Xb = function g() = 2
contract Cntr =
  using Xa as A
  using Xb as A
  type state = int
  entrypoint init() = A.f() + A.g()
```

I think the above example should compile, and I can't think of a way that this could cause any bugs (beside that, haskell allows using the same alias for multiple modules).

hanssv:
Again, let me remind me about the target audience... Modeling things after Haskell might or might not be the correct thing 😅 

That last example did not match my mental picture of things but I should be able to adapt the picture - possibly the documentation could clarify this?

hanssv (dismissed):

ghallak:
@hanssv I've updated the docs to here 7fabbff11e07da8451df3d57a281f9bae7b3d739

hanssv (approved):
👍

Issue #232, opened 2020-02-20
Pattern guards


radrow:
Sometimes it is nice to be able to provide some assertions on patterns, for instance:
```
function f(l) = switch(l)
  k::_ | k < 0 => k
  _::t         => f(t)
  []           => 0
```
Expressions in the guard must not be stateful

PR #339, opened 2021-09-02
Pattern guards for functions and switch statements

ghallak:
Fixes: #232


docs/sophia_features.md
```diff
@@ -482,6 +482,25 @@ function g(p : int * option(int)) : int =
   b
 ```
 
+Guards are boolean functions that can be used on patterns in both switch
```
radrow, line 485: I'd elaborate a little bit on that. Note that many smart contract devs aren't familiar with functional programming. Also, mention that these expressions cannot be stateful

hanssv:
Looks good from a quick look, are there any (actually running) tests over in `aeternity` repo?

src/aeso_syntax.erl
```diff
@@ -57,7 +57,8 @@
 -type pragma() :: {compiler, '==' | '<' | '>' | '=<' | '>=', compiler_version()}.
 
 -type letval()  :: {letval, ann(), pat(), expr()}.
--type letfun()  :: {letfun, ann(), id(), [pat()], type(), expr()}.
+-type letfun()  :: {letfun, ann(), id(), [pat()], type(), expr()}
+                 | {letfun, ann(), id(), [pat()], type(), expr(), expr()}.
```
radrow, line 61: For the user facing syntax I think it would be better to have just one variant for simplicity. If there are no guards, then the expr could be `true` which can be optimized out later on. Also, I would rather opt for comma separated multiple guards -- this would ease implementing assigning values in guards like you can do in Haskell (but that's for another task). Then, no guards would be just an empty list here. @UlfNorell what do you think?

src/aeso_ast_infer_types.erl
```diff
@@ -1366,13 +1391,25 @@ infer_letfun1(Env0, {letfun, Attrib, Fun = {id, NameAttrib, Name}, Args, What, B
     NamedArgs = [],
     TypeSig = {type_sig, Attrib, none, NamedArgs, ArgTypes, ResultType},
     {{Name, TypeSig},
-     {letfun, Attrib, {id, NameAttrib, Name}, TypedArgs, ResultType, NewBody}}.
+     {letfun, Attrib, {id, NameAttrib, Name}, TypedArgs, ResultType, NewBody}};
+infer_letfun1(Env0, {letfun, Attrib, Fun = {id, NameAttrib, Name}, Args, What, Guard, Body}) ->
+    Env = Env0#env{ stateful = aeso_syntax:get_ann(stateful, Attrib, false),
+                    current_function = Fun },
+    {NewEnv, {typed, _, {tuple, _, TypedArgs}, {tuple_t, _, ArgTypes}}} = infer_pattern(Env, {tuple, [{origin, system} | NameAttrib], Args}),
+    NewGuard = check_expr(NewEnv, Guard, {id, Attrib, "bool"}),
```
radrow, line 1399: Remove stateful context so there are no surprises

radrow:
Add some negative test maybe?

src/aeso_syntax.erl
UlfNorell: Not introducing redundant forms in the AST is good, and comma separated guards make sense. Especially if we want to add support for what is called pattern guards in Haskell land later. (What you have here is referred to as just *guards*.)

src/aeso_ast_infer_types.erl
ghallak: Done here 756c05c60038d774ce1156c7b65896ea9741036e

docs/sophia_features.md
ghallak: Let me know if you think this makes it clearer 2deb568be3081db48f2ea0ddf77337aa1bde88ba

src/aeso_syntax.erl
```diff
@@ -145,7 +146,8 @@
 -type stmt() :: letbind()
               | expr().
 
--type alt() :: {'case', ann(), pat(), expr()}.
+-type alt() :: {'case', ann(), pat(), expr()}
+             | {'case', ann(), pat(), expr(), expr()}.
```
radrow, line 150:
```suggestion
-type alt() :: {'case', ann(), pat(), [guard()], expr()}.

-type guard() :: expr().
```
So we may add assign-guards later on here

src/aeso_parser.erl
```diff
@@ -211,12 +211,20 @@ letdef() -> choice(valdef(), fundef()).
...
+unguarded_fundef() ->
     choice(
     [ ?RULE(id(), args(),                   tok('='), body(), {letfun, get_ann(_1), _1, _2, type_wildcard(get_ann(_1)), _4})
     , ?RULE(id(), args(), tok(':'), type(), tok('='), body(), {letfun, get_ann(_1), _1, _2, _4, _6})
     ]).
 
+guarded_fundef() ->
+    choice(
+    [ ?RULE(id(), args(),                   tok('|'), expr(), tok('='), body(), {letfun, get_ann(_1), _1, _2, type_wildcard(get_ann(_1)), _4, _6})
+    , ?RULE(id(), args(), tok(':'), type(), tok('|'), expr(), tok('='), body(), {letfun, get_ann(_1), _1, _2, _4, _6, _8})
```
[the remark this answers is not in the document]
radrow: Might be also good to make `guard/0` a separate parser

src/aeso_parser.erl
```diff
@@ -211,12 +211,20 @@ letdef() -> choice(valdef(), fundef()).
 valdef() ->
     ?RULE(pattern(), tok('='), body(), {letval, [], _1, _3}).
 
-fundef() ->
+fundef() -> choice(unguarded_fundef(), guarded_fundef()).
```
radrow, line 214: Isn't there some combinator for optional parsing?

src/aeso_syntax.erl
```diff
@@ -146,8 +145,7 @@
 -type stmt() :: letbind()
               | expr().
 
--type alt() :: {'case', ann(), pat(), expr()}
-             | {'case', ann(), pat(), expr(), expr()}.
+-type alt() :: {'case', ann(), pat(), [expr()], [expr()]}.
```
radrow, line 148: why list of expr?

ghallak: multiple guards, each guard corresponds to one body.

ghallak: multiple guards, each guard corresponds to one body.

radrow: I don't get it, how do you represent
```
switch(x)
  (a, b) | a > 0, b > 0 => a + b
         | true => 0 
```
?

radrow: Or if it is invalid
```
switch(x)
  (a, b) | a > 0, b > 0 => a + b
  _  => 0 
```

ghallak: The above does is not valid since the guard should be a single expr, so if you write `a > 0 && b > 0` instead of `a > 0, b > 0` in the first example it would be parsed like this:

```
[{switch,                                                                                     [3/4542]
     [{file,no_file},{line,4},{col,5}],
     {id,[{file,no_file},{line,4},{col,12}],"x"},
     [{'case',
          [{file,no_file},{line,6},{col,26}],
          {tuple,
              [{file,no_file},{line,5},{col,7}],
              [{id,[{file,no_file},{line,5},{col,8}],"a"},
               {id,[{file,no_file},{line,5},{col,11}],"b"}]},
          [{app,
               [{file,no_file},{line,6},{col,11},{format,infix}],
               {'&&',[{file,no_file},{line,6},{col,17}]},
               [{app,
                    [{file,no_file},{line,6},{col,11},{format,infix}],
                    {'>',[{file,no_file},{line,6},{col,13}]},
                    [{id,[{file,no_file},{line,6},{col,11}],"a"},
                     {int,[{file,no_file},{line,6},{col,15}],0}]},
                {app,
                    [{file,no_file},{line,6},{col,20},{format,infix}],
                    {'>',[{file,no_file},{line,6},{col,22}]},
                    [{id,[{file,no_file},{line,6},{col,20}],"b"},
                     {int,[{file,no_file},{line,6},{col,24}],0}]}]},
           {bool,[{file,no_file},{line,7},{col,11}],true}],
          [{app,
               [{file,no_file},{line,6},{col,29},{format,infix}],
               {'+',[{file,no_file},{line,6},{col,31}]},
               [{id,[{file,no_file},{line,6},{col,29}],"a"},
                {id,[{file,no_file},{line,6},{col,33}],"b"}]},
           {int,[{file,no_file},{line,7},{col,19}],0}]}]}]}]}]
```

Notice that there are 2 guards and 2 bodies.

Do you think there's something wrong with that?

docs/sophia_features.md
ghallak: Done that here 70af4828483be4a8ac93a9b123411e2e0b946fef

src/aeso_parser.erl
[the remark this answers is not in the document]
ghallak: Why do you think that guards should have comma separated exprs? Isn't it better if a single expr is used and boolean operators are used when needed?

src/aeso_syntax.erl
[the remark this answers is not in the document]
ghallak: If there are multiple guards and a single body, how would you parse something like

```
switch(x)
  (a, b) | a > 0, b > 0 => a + b
         | true => 0 
```

src/aeso_syntax.erl
radrow: Uh, when talking about multiple guards we were talking about `| a > 0, b > 0`, so in the future we might have `| a > 0, b > 0, let x = a + b, x > 10`. That's why there would be a list of guards.

Multiple guard alternatives for a single pattern are great too, but then why not pack them into a list of tuples? They are conceptually together, so imo they should stay side by side. 

My idea for the typing:

```erlang
-type alt() :: {'case', ann(), pat(), [guard_alt()]}.
-type guard_alt :: {guarded, ann(), [guard()], expr()}.
-type guard() :: expr().
```

This way this:
```
switch(x)
  (a, b) | a > 0, b > 0 => a + b
         | true => 0 
```
can be expressed as (simplified for readability)
```
{switch, _, "x", [
  {case, _, {tuple, ["a", "b"]}, [
    {guarded, _, [a > 0, b > 0], a + b},
    {guarded, _, true, 0}
  ]}
]}
```

src/aeso_parser.erl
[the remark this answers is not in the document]
radrow: Because later we might want to add support for introducing variables in guards. Just like in Haskell you can do
```
f x | y <- x + x, y > 10 = 10
```

ghallak:
> Looks good from a quick look, are there any (actually running) tests over in `aeternity` repo?

There are tests here now: https://github.com/ghallak/aeternity/commit/4683da5b500dd764f62d4c8b4f3f1e83e9723fd7

docs/sophia_features.md
```diff
@@ -482,7 +484,28 @@ function g(p : int * option(int)) : int =
...
+function get_left_if_positive(x : one_or_both('a, 'b)) : option('a) =
+  switch(x)
+    Left(x)    | x > 0 => Some(x)
+    Both(x, _) | x > 0 => Some(x)
+    _                  => None
+```
+
+```sophia
+function
+  get_left_if_positive : one_or_both('a, 'b) => option('a)
```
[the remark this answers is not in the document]
ghallak: Now it does 91002c1f77fc164484f05c1ab82708805c5f35bb

test/aeso_abi_tests.erl
```diff
@@ -56,36 +56,37 @@ encode_decode_test() ->
...
-    ?assertEqual("Data error:\nFailed to decode binary as type string\n", aeso_errors:pp(Err1)),
-    {error, [Err2]} = aeso_compiler:to_sophia_value(Code, "x", ok, encode(12), [{backend, fate}]),
-    ?assertEqual("Data error:\nFailed to decode binary as type string\n", aeso_errors:pp(Err2)),
-
-    {error, [Err3]} = aeso_compiler:to_sophia_value(Code, "x", revert, encode(12)),
-    ?assertEqual("Data error:\nCould not interpret the revert message\n", aeso_errors:pp(Err3)),
-    {error, [Err4]} = aeso_compiler:to_sophia_value(Code, "x", revert, encode(12), [{backend, fate}]),
-    ?assertEqual("Data error:\nCould not deserialize the revert message\n", aeso_errors:pp(Err4)),
-    ok.
+encode_decode_sophia_test() -> ok.
```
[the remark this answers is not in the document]
ghallak: In this commit e828099bd97dffe11438f2e48f3a92ce3641e85b I have disabled multiple tests that are related to aevm, including this one.

To make these tests work, changes are required to be made to icode generation which is deprecated as I understood.

Should I make these changes to icode generation? or is there another way to re-enable these tests?

PR #346, opened 2021-10-04
Add compiler warnings

ghallak:
Currently added warnings:

- [x] shadowing
- [x] negative spends
- [x] division by zero
- [x] unused functions
- [x] unused includes
- [x] unused `stateful` annotations
- [x] unused variables
- [x] unused parameters
- [x] unused user-defined type
- [x] dead return value
- [ ] ~non-exhaustive patterns~ (this will be tricky and should better be done in a separate PR)

Currently supported options:

- [x] Turn options off/on separately.
- [x] Warnings packages (Wall). pedantic package can easily be added later.
- [x] Turn warnings into type errors.

Things that still need to be done:
- [x] Properly report the warnings on compilation. (Currently they are just stored but not reported)

Testing:
- [x] Tests for all added warnings
- [x] Test for turning warnings into type errors

### DIFF
--- a/src/aeso_ast_infer_types.erl
+++ b/src/aeso_ast_infer_types.erl
@@ -171,7 +171,8 @@ on_scopes(Env = #env{ scopes = Scopes }, Fun) ->
     Env#env{ scopes = maps:map(fun(_, Scope) -> Fun(Scope) end, Scopes) }.
 
 -spec bind_var(aeso_syntax:id(), utype(), env()) -> env().
-bind_var({id, Ann, X}, T, Env) ->
+bind_var({id, Ann, X}, T, Env = #env{ vars = Vars }) ->
+    when_warning(warn_shadowing, fun() -> warn_potential_shadowing(Ann, X, Vars) end),
     Env#env{ vars = [{X, {Ann, T}} | Env#env.vars] }.
 
 -spec bind_vars([{aeso_syntax:id(), utype()}], env()) -> env().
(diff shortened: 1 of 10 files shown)


src/aeso_ast_infer_types.erl
```diff
@@ -2869,6 +2870,7 @@ destroy_and_report_unused_functions() ->
 
 %% Warnings (Shadowing)
 
+warn_potential_shadowing(_, "_", _) -> ok;
```
dincho, line 2882: _"_"_ 🤷🏻‍♂️

ghallak:
For negative spends and division by zero, the warning will only show when using a literal.

```
function f() =
  let x = 1 / 0  // this is a warning
  
  let zero = 0
  let y = 1 / zero  // no warning will show here

  let z = 1 / (1 - 1)  // no warning will show here
```

I think it won't be very difficult to add warning for the second case `1 / zero` since this will only need to lookup the value of the variables `zero`, but the third case will requires evaluating the expression `1 - 1`, which will make things more complicated and I'm not sure if I should try to add that.

dincho:
> but the third case will requires evaluating the expression `1 - 1`, which will make things more complicated and I'm not sure if I should try to add that.

I guess if the compiler have a optimization phase to evaluate that and you run the warning check in a step/phase after it it would be covered by case 1

radrow:
there could be a second pass of warning check after additional constant propagation

hanssv (approved):
👍 Looks good!

radrow (approved):

radrow:
Would be nice to reflect it in the cli and http interface too

radrow:
@nikita-fuchs would be great to see it working in aestudio :)

nikita-fuchs:
> @nikita-fuchs would be great to see it working in aestudio :)

dayumn, thanks for bringing this up.

Issue #337, opened 2021-08-13
bit shifting over numbers


bogdan-manole:
I'm trying to implement some similar contracts from uniswap into aesophia. There are a lot of bitwise operations there, like: left shift , right shift over integer types. I think it will be helpful if we could have functions as shl, shr, not over bits type and also some way to convert between int and bits.

hanssv:
Adding built in bitwise operations in FATE/Sophia needs a protocol upgrade (a.k.a a hard fork) - in the meantime we add a standard library: #340

hanssv:
Merged to the Ceres branch.

PR #352, opened 2021-11-04
Add bitwise operations, MULMOD, addr_to_bytes and Poseidon hash

hanssv:
These are extensions to Sophia (and eventually FATE) for the Ceres protocol update. It covers #337 and more.


docs/sophia_syntax.md
```diff
@@ -255,9 +257,14 @@ In order of highest to lowest precedence.
 | `!` | right
 | `^` | left
 | `*` `/` `mod` | left
+| `bnot` | right
```
UlfNorell, line 260: I would put `bnot` at the same level as `!`. This is what both C and Erlang does.

UlfNorell (changes requested):

UlfNorell (approved):

radrow (approved):

Issue #307, opened 2021-04-14
Polymorphism support (Must implement function if included an interface)


genievot:
Greetings, 

I was wondering if there is a way to implement interfaces in Sophia so we can work on standards in a more appropriate way. As discussed here https://forum.aeternity.com/t/how-to-write-interfaces-in-sophia/8236/6 .

Best,

radrow:
I see it as something like 
```
contract interface Strokable =
  entrypoint stroke : () => unit

contract Cat : Strokable =
  @impl 
  entrypoint stroke() = abort("not today")
```
or even
```
contract Cat =
  impl Strokable =
    entrypoint stroke() = abort("not today")
  entrypoint ...
```

The question is how deep do we want to dive into it:
- Do we want to let interfaces be subinterfaces of the others (smell Java)
- What about multiple inheritance?
- Should the type system treat it in Rust/Haskellish way as a typeclass or in OOP manner with subtyping?

genievot:
Thank you for looking into it, My view on your questions:

-> I was not sure about this but could be helpful in the case of extended standards (Including a function that was not in the original interface, for e.g. adding `safeTransfer` if not in default ERC721) so should result in a cleaner code & better understanding.
-> Not sure If I currently have any use of this either but in the future, this might be helpful in writing cleaner modules (classes) and not deriving every class before moving on (Since you asked so I assume that currently, Sophia does not support multiple inheritance).
-> Not sure about this but whatever you prefer I can adapt to it.

Best,

marc0olo:
I personally like this kind of style better:
```
contract interface Strokable =
  entrypoint stroke : () => unit

contract Cat : Strokable =
  @impl 
  entrypoint stroke() = abort("not today")
```

whereas the compilation should fail if the implementation of `stroke()` is missing in the contract.

- we should definitely provide multiple inheritance for interfaces
- subinterfaces would be great but I don't know whether this is a bit contrary to the current state of Sophia where we don't have contract inheritance (https://github.com/aeternity/aesophia/issues/198)
   - this should probably be introduced together
- I have no preference in regards to the type system, but probably the rust/haskell way fits Sophia better

arpunk:
I might be biased, but what about:

```
interface Strokable =
  callback stroke : () => unit

interface Other =
  callback other : () => unit

contract Cat =
  @impl Strokable
  entrypoint stroke() = abort("not today")
  
  @impl Other
  entrypoint other() = about("probably tomorrow")
```

[...]

PR #357, opened 2021-11-24
Polymorphism support

ghallak:
Fixes: #307


ghallak:
I have noticed that the type for the function declaration `entrypoint init : () => unit` is:
```
{fun_t,_,[],[],{id,_,"unit"}}
```

While the type of the function `entrypoint init() = ()` is:
```
{tuple_t,_,[]}
```

The function `aeso_ast_infer_types:unfold_types_in_type` replaces `{id, _, "unit"}` with `{tuple_t, _, []}`, but it does not seems to be called on the `fun_t` type.

@hanssv Do you know if this is a bug or if it's done on purpose?

hanssv:
I have no recollection of it being done on purpose...

src/aeso_aci.erl
```diff
@@ -83,7 +83,7 @@ from_typed_ast(Type, TypedAst) ->
         string -> do_render_aci_json(JArray)
     end.
 
-encode_contract(Contract = {Head, _, {con, _, Name}, _}) when ?IS_CONTRACT_HEAD(Head) ->
```
hanssv, line 86: I think we are soon approaching the point where this should become a record, something to consider for the _next_ addition 😅

src/aeso_parser.erl
```diff
@@ -96,17 +96,29 @@ decl() ->
     choice(
       %% Contract declaration
     [ ?RULE(token(main), keyword(contract),
-            con(), tok('='), maybe_block(decl()), {contract_main, _2, _3, _5})
+            con(), tok('='), maybe_block(decl()), {contract_main, _2, _3, [], _5})
```
hanssv, line 99: This looks like an exponential blowup - we should consider a rewrite some day...

hanssv (approved):
Looks good to me 👍

hanssv (dismissed):

radrow (changes requested):
I would like to see tests for:

- Implementing many interfaces
- Implementing two interfaces with entrypoints of same names and types
- Implementing two interfaces with entrypoints of same names and different types
- Variance switching case, in particular that assuming `contract Cat : Animal`:
  - `let f(x : Cat) : Animal = (g : Animal => Cat)` works
  - `let f(x : Animal) : Animal = (g : Cat=> Cat)` does not work
  - `let f(x : Animal => Cat) : Animal = (g : (Cat => Animal) => Cat)` works
  - `let f(x : Animal => Cat) : Animal = (g : (Animal => Animal) => Cat)` does not work
  - `let f(x : Animal => Cat) : Animal = (g : (Cat => Cat ) => Cat)` does not work
  - `let f(x : Animal => Animal ) : Animal = (g : (Cat => Animal) => Cat)` does not work
  - `let f(x : Cat => Cat) : Animal = (g : (Cat => Animal) => Cat)` does not work

radrow:
Also, please refer some runtime tests in aeternity repo, so we are sure nothing breaks there.

ghallak:
@hanssv @UlfNorell  In the case of implementing two interfaces with entrypoints of the same name and same type

```
contract interface I =
    entrypoint f : () => int

contract interface J =
    entrypoint f : () => int

contract C : I, J =
    entrypoint f() = 1
```

Do you think this should pass as a function called `f` with the type `() => int` is already implemented? Or should it fail since only one of the interfaces is implemented, and not the other?

I already have [this test](https://github.com/aeternity/aesophia/blob/8c42f69407778e946fd0ba2dff400bc941a2721b/test/contracts/contract_polymorphism_same_decl_multi_interface.aes) and it's failing, but I'm wondering if it should.

UlfNorell:
I think the least error prone option is to not allow implementing interfaces with overlapping entrypoints. So in this case it should give an error already on `contract C : I, J =`. I don't see any great harm in allowing it though, so if there are convincing use cases where you want overlapping interfaces we could accept it.

test/contracts/contract_interface_polymorphism.aes
```diff
@@ -0,0 +1,9 @@
+contract interface II =
```
radrow, line 1: Do we have a test if C does not implement f?

ghallak: Yes, [contract_polymorphism_missing_implementation](https://github.com/aeternity/aesophia/blob/8c42f69407778e946fd0ba2dff400bc941a2721b/test/contracts/contract_polymorphism_missing_implementation.aes) and it shows an error.

marc0olo:
@ghallak I agree with Ulf, as long as there is no valid usecase to argue for we shouldn't allow implementing interfaces with overlapping entrypoints

radrow:
I'd also modify some tests to involve tuples, maps, lists etc, so we are sure that subtyping works on them too

src/aeso_ast_infer_types.erl
```diff
@@ -2556,55 +2694,94 @@ unify1(_Env, {uvar, A, R}, T, When) ->
...
-unify1(_Env, {fun_t, _, _, var_args, _}, {fun_t, _, _, _, _}, When) ->
+unify1(_Env, {fun_t, _, _, var_args, _}, {fun_t, _, _, _, _}, _Variance, When) ->
     type_error({unify_varargs, When});
-unify1(Env, {fun_t, _, Named1, Args1, Result1}, {fun_t, _, Named2, Args2, Result2}, When)
+unify1(Env, {fun_t, _, Named1, Args1, Result1}, {fun_t, _, Named2, Args2, Result2}, Variance, When)
   when length(Args1) == length(Args2) ->
-    unify(Env, Named1, Named2, When) andalso
-    unify(Env, Args1, Args2, When) andalso unify(Env, Result1, Result2, When);
-unify1(Env, {app_t, _, {Tag, _, F}, Args1}, {app_t, _, {Tag, _, F}, Args2}, When)
+    OppositeVariance = case Variance of
```
radrow, line 2738: this could be a separate function imo

src/aeso_ast_infer_types.erl
```diff
@@ -2556,55 +2694,94 @@ unify1(_Env, {uvar, A, R}, T, When) ->
...
     true;
-unify1(_Env, {con, _, Name}, {con, _, Name}, _When) ->
-    true;
-unify1(_Env, {qid, _, Name}, {qid, _, Name}, _When) ->
+unify1(Env, A = {con, _, NameA}, B = {con, _, NameB}, Variance, When) ->
+    IsSubtype = case Variance of
+                    invariant     -> NameA == NameB;
+                    covariant     -> is_subtype(Env, NameA, NameB);
+                    contravariant -> is_subtype(Env, NameB, NameA);
+                    bivariant     -> is_subtype(Env, NameA, NameB)
```
radrow, line 2713: I guess?
```suggestion
                    bivariant     -> is_subtype(Env, NameA, NameB) orelse is_subtype(Env, NameB, NameA)
```

src/aeso_ast_infer_types.erl
```diff
@@ -2556,55 +2694,94 @@ unify1(_Env, {uvar, A, R}, T, When) ->
...
+unify1(_Env, X, X, _Variance, _When) ->
     true;
-unify1(_Env, {id, _, Name}, {id, _, Name}, _When) ->
+unify1(_Env, {id, _, Name}, {id, _, Name}, _Variance, _When) ->
     true;
-unify1(_Env, {con, _, Name}, {con, _, Name}, _When) ->
-    true;
-unify1(_Env, {qid, _, Name}, {qid, _, Name}, _When) ->
+unify1(Env, A = {con, _, NameA}, B = {con, _, NameB}, Variance, When) ->
+    IsSubtype = case Variance of
```
radrow, line 2709: This could be a `is_subtype/4` that calls `is_subtype/3`, less indentation

src/aeso_parser.erl
radrow: @ghallak would you leave a TODO comment to verify it?

radrow: or just check if it has a performance danger

src/aeso_ast_infer_types.erl
ghallak: Done here 7b5c9778b4f39dd6a96c1ea07d5978031ca90bc8

src/aeso_ast_infer_types.erl
ghallak: Done here cd3da44973ca25574079ae4f1c1a930c8828d6d6

src/aeso_ast_infer_types.erl
ghallak: Done here cd3da44973ca25574079ae4f1c1a930c8828d6d6

src/aeso_ast_infer_types.erl
```diff
@@ -1024,6 +1104,8 @@ check_typedef_sccs(Env, TypeMap, [{acyclic, Name} | SCCs], Acc) ->
                     Env2 = check_fields(Env1, TypeMap, RecTy, Fields),
                     check_typedef_sccs(Env2, TypeMap, SCCs, Acc1);
                 {variant_t, Cons} ->
+                    ets_insert(type_vars_variance, {Env#env.namespace ++ qname(D),
```
radrow, line 1107: What does it mean here?

src/aeso_ast_infer_types.erl
```diff
@@ -1049,6 +1131,73 @@ check_typedef(Env, {variant_t, Cons}) ->
...
+    FlatArgs = lists:flatten([Args || {constr_t, _, _, Args} <- Cons]),
+
+    Vs = lists:flatten([infer_type_vars_variance(Arg) || Arg <- FlatArgs]),
+    lists:map(fun({tvar, _, TVar}) ->
+                      S = sets:from_list([Variance || {TV, Variance} <- Vs, TV == TVar]),
+                      IsInvariant     = sets:is_element(invariant, S),
+                      IsCovariant     = sets:is_element(covariant, S),
+                      IsContravariant = sets:is_element(contravariant, S),
+                      IsBivariant     = sets:is_element(bivariant, S),
+                      case {IsInvariant, IsCovariant, IsContravariant, IsBivariant} of
```
radrow, line 1145: I think you can just have a set for co and contra and then 
```
case {Co, Contra} of
  {t, f} -> co
  {f, t} -> contra
  {t, t} -> inv
  {f, f} -> bi
```

src/aeso_ast_infer_types.erl
```diff
@@ -1049,6 +1131,73 @@ check_typedef(Env, {variant_t, Cons}) ->
...
+              1 -> {TVar, contravariant}
+          end,
+    [Cur | infer_type_vars_variance(Res)];
+infer_type_vars_variance({fun_t, _, [], [_Arg], Res}) ->
+    infer_type_vars_variance(Res);
+infer_type_vars_variance({tvar, _, TVar}) ->
+    [{TVar, covariant}];
+infer_type_vars_variance(_) -> [].
+
+arrows_in_type({fun_t, _, [], [_Arg], FRes}) ->
```
radrow, line 1191: I think it would be easier and simpler to use `infer_type_vars_variance` to dive into the type and just maintain an additional argument that tells what the current variance is. Then you won't need to count this.

test/aeso_compiler_tests.erl
```diff
@@ -834,6 +837,173 @@ failing_contracts() ->
...
+                   <<?Pos(50,55)
+                     "Cannot unify `Animal` and `Cat`\n"
+                     "when checking the application of\n"
+                     "  `g5 : ((Animal) => Animal) => Cat`\n"
+                     "to arguments\n"
+                     "  `x : (Cat) => Cat`">>
+                  ])
+    , ?TYPE_ERROR(polymorphism_variance_switching_custom_types,
+                  [<<?Pos(56,32)
+                     "Cannot unify `Cat` and `Animal`\n"
```
radrow, line 899: The error messages here could be more elaborate. I can already see people scratching their heads why Cat does not unify with Animal, but sometimes the other way around. I think it would be valuable to mention the variance in the message:
```
Cannot unify A with B in a contravariant context
```
Or something. Even if one does not know the concept of variance, they know at least what to google.

test/aeso_compiler_tests.erl
```diff
@@ -218,6 +218,9 @@ compilable_contracts() ->
      "using_namespace",
      "assign_patterns",
      "patterns_guards",
+     "contract_polymorphism",
+     "contract_polymorphism_multi_interface",
+     "contract_interface_polymorphism",
```
radrow, line 223: Add tests for custom datatypes, records and type aliases. Please consider various variances. Also there should be tests for tuples, maps, lists and oracles.

src/aeso_ast_infer_types.erl
```diff
@@ -846,7 +854,25 @@ infer1(Env, [{Contract, Ann, ConName, Code} | Rest], Acc, Options)
...
+                               end
+                  end, Impls),
+    destroy_and_report_type_errors(Env),
+    create_type_errors(),
+    case What of
+        contract ->
+            ImplementedInterfaces = [proplists:get_value(Name, AllInterfaces) || Name <- ImplsNames],
+            check_implemented_interfaces(ImplementedInterfaces, ConName, [ Fun || Fun = {letfun, _, _, _, _, _} <- Code1 ], [], AllInterfaces);
+        contract_interface ->
+            ok
```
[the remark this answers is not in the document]
radrow: Note that they serve for remote contracts as well!

src/aeso_ast_infer_types.erl
```diff
@@ -1024,6 +1104,8 @@ check_typedef_sccs(Env, TypeMap, [{acyclic, Name} | SCCs], Acc) ->
                     Env2 = check_fields(Env1, TypeMap, RecTy, Fields),
                     check_typedef_sccs(Env2, TypeMap, SCCs, Acc1);
                 {variant_t, Cons} ->
+                    ets_insert(type_vars_variance, {Env#env.namespace ++ qname(D),
+                                                    infer_type_vars_variance(Xs, Cons)}),
                     Target   = check_type(Env1, app_t(Ann, D, Xs)),
```
radrow, line 1109: I make this note here because I couldn't find any better place; what about builtin types such as option, oracle and oracle_query? How do you handle them?

src/aeso_ast_infer_types.erl
```diff
@@ -846,7 +854,25 @@ infer1(Env, [{Contract, Ann, ConName, Code} | Rest], Acc, Options)
...
+    Contract1 = {Contract, Ann, ConName, Impls, Code1},
+    AllInterfaces = [{name(IName), I} || I = {contract_interface, _, IName, _, _} <- Acc],
+    ImplsNames = lists:map(fun name/1, Impls),
+    create_type_errors(),
+    lists:foreach(fun(Impl) -> case proplists:get_value(name(Impl), AllInterfaces) of
+                                   undefined -> type_error({referencing_undefined_interface, Impl});
+                                   _         -> ok
+                               end
+                  end, Impls),
+    destroy_and_report_type_errors(Env),
```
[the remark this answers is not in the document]
ghallak: Removing that was causing a failed test, but I've just found out that the failing test was failing because of another mistake that I've made.

Fixed here: 4790566f427a3f9ec01b41a61217a7fe319d2525

src/aeso_ast_infer_types.erl
```diff
@@ -846,7 +854,25 @@ infer1(Env, [{Contract, Ann, ConName, Code} | Rest], Acc, Options)
         contract_interface -> ok
     end,
     {Env1, Code1} = infer_contract_top(push_scope(contract, ConName, Env), What, Code, Options),
-    Contract1 = {Contract, Ann, ConName, Code1},
+    Contract1 = {Contract, Ann, ConName, Impls, Code1},
+    AllInterfaces = [{name(IName), I} || I = {contract_interface, _, IName, _, _} <- Acc],
+    ImplsNames = lists:map(fun name/1, Impls),
+    create_type_errors(),
+    lists:foreach(fun(Impl) -> case proplists:get_value(name(Impl), AllInterfaces) of
```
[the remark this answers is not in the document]
ghallak: Done here 0a2e11e4e69b68ab10ee1f01e10b776172379664

src/aeso_ast_infer_types.erl
```diff
@@ -846,7 +854,25 @@ infer1(Env, [{Contract, Ann, ConName, Code} | Rest], Acc, Options)
...
+                                   undefined -> type_error({referencing_undefined_interface, Impl});
+                                   _         -> ok
+                               end
+                  end, Impls),
+    destroy_and_report_type_errors(Env),
+    create_type_errors(),
+    case What of
```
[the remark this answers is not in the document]
ghallak: Done here 0a2e11e4e69b68ab10ee1f01e10b776172379664

src/aeso_ast_infer_types.erl
```diff
@@ -862,17 +888,71 @@ infer1(Env, [{pragma, _, _} | Rest], Acc, Options) ->
     %% Pragmas are checked in check_modifiers
     infer1(Env, Rest, Acc, Options).
 
+check_implemented_interfaces([], _, _, _, _) ->
```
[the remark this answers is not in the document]
ghallak: Done here 0a2e11e4e69b68ab10ee1f01e10b776172379664

src/aeso_ast_infer_types.erl
```diff
@@ -862,17 +888,71 @@ infer1(Env, [{pragma, _, _} | Rest], Acc, Options) ->
...
+    case lists:member(name(IName), Acc) of
+        true ->
+            check_implemented_interfaces(Interfaces, ConId, Impls, Acc, AllInterfaces);
+        false ->
+            Unmatched = match_impls(Decls, ConId, name(IName), Impls),
+            NewInterfaces = [proplists:get_value(name(I), AllInterfaces) || I <- Extensions],
+            check_implemented_interfaces(Interfaces ++ NewInterfaces, ConId, Unmatched, [name(IName) | Acc], AllInterfaces)
+    end.
+
+match_impls([], _, _, Impls) ->
```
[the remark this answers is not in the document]
ghallak: Done here 0a2e11e4e69b68ab10ee1f01e10b776172379664

src/aeso_ast_infer_types.erl
```diff
@@ -862,17 +888,71 @@ infer1(Env, [{pragma, _, _} | Rest], Acc, Options) ->
     %% Pragmas are checked in check_modifiers
     infer1(Env, Rest, Acc, Options).
 
+check_implemented_interfaces([], _, _, _, _) ->
+    ok;
+check_implemented_interfaces([{contract_interface, _, IName, Extensions, Decls} | Interfaces], ConId, Impls, Acc, AllInterfaces) ->
+    case lists:member(name(IName), Acc) of
+        true ->
+            check_implemented_interfaces(Interfaces, ConId, Impls, Acc, AllInterfaces);
```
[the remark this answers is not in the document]
ghallak: There is a comment in this commit https://github.com/aeternity/aesophia/commit/0a2e11e4e69b68ab10ee1f01e10b776172379664 explaining what the function is doing.

src/aeso_ast_infer_types.erl
```diff
@@ -862,17 +888,71 @@ infer1(Env, [{pragma, _, _} | Rest], Acc, Options) ->
...
+                             lists:delete(V, Impls);
+                         false ->
+                             type_error({unimplemented_interface_function, ConId, IName, FunName}),
+                             Impls
+                     end,
+    match_impls(Decls, ConId, IName, UnmatchedImpls).
+
+
+-spec compare_types(T, T) -> boolean() when T :: utype() | [utype()].
+compare_types(Types1 = [_ | _], Types2 = [_ | _]) ->
```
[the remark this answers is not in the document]
ghallak: It's been a long time since I wrote this. I didn't know about `is_list/1` back then and thought this is the only way to check if a variable is a list.

Fixed here: 55bc58286e224ab54fff62a3b8ae2cd0ca87adcb

src/aeso_ast_infer_types.erl
```diff
@@ -862,17 +888,71 @@ infer1(Env, [{pragma, _, _} | Rest], Acc, Options) ->
...
+    length(Types1) == length(Types2) andalso
+    lists:all(fun({T1, T2}) -> compare_types(T1, T2) end, lists:zip(Types1, Types2));
+compare_types({fun_t, _, _, Types1, RetType1}, {fun_t, _, _, Types2, RetType2}) ->
+    % TODO: what about named_args_t and var_args?
```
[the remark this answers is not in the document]
ghallak: Done here 62b095affc17a12ff634da242e212c51cbd36a99

src/aeso_ast_infer_types.erl
ghallak: Done here bdd3f0dde2f4423722808cdd5478a7a54d62f85f

src/aeso_ast_infer_types.erl
ghallak: Done here 39b9bb66cca29f836f8c36a3569df494dbd19aff

src/aeso_ast_infer_types.erl
```diff
@@ -1639,6 +1788,16 @@ infer_expr(Env, {app, Ann, Fun, Args0} = App) ->
             ResultType = fresh_uvar(Ann),
             when_warning(warn_unused_functions,
                          fun() -> register_function_call(Namespace ++ qname(CurrentFun), Name) end),
+            % the uvars of tvars are stored so that no variance switching happens
+            % in between them (e.g. in TC('a => 'a), 'a should be a single type)
+            case FunType of
```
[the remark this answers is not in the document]
ghallak: This is supposed to work in the case where the function is a type constructor. So it would work for something like:

```
let x = T(f)
```

where `f` is defined as:

```
function f(_ : Animal) : Cat = some_cat
```

and `T` is a type constructor for the datatype `tc`:

```
datatype tc('a) = T('a => 'a)
```

In the above example, when trying to infer the type for `T(f)`, the `FunType` for `T(f)` would look something like:

```
{fun_t,
    Ann1,
    [],
    [{fun_t, Ann2, [], [{uvar, Ann3, Ref}], {uvar, Ann4, Ref}}],
    {app_t, Ann5, {qid, Ann6, ["Main", "tc"]}, [{uvar, Ann7, Ref}]}}
```

Note that `Ref` in the return type is the same as the `Ref` from the args, since the type above is `    datatype tc('a) = TC('a => 'a)`, so it's enough to keep track of the `Ref` in the return type.

test/aeso_compiler_tests.erl
ghallak: Would it be better to say that `Animal` is not a subtype of `Cat` instead of mentioning the variance part? Someone would scratch his head saying that why didn't the compiler check if `Cat` is a subtype of `Animal` instead of checking if `Animal` is a subtype of `Cat, but people will always scratch their heads for some reason, and I don't see that mentioning the type of variance making it easier for people to find out what went wrong.

src/aeso_ast_infer_types.erl
ghallak: I probably don't get the question.

Here we have access to the datatype's type variables `Xs` and the constructors for the datatype `Cons`, and I use that to infer the type variance for these type variables.

Are you asking about something else?

src/aeso_ast_infer_types.erl
```diff
@@ -862,17 +888,71 @@ infer1(Env, [{pragma, _, _} | Rest], Acc, Options) ->
...
+                             type_error({unimplemented_interface_function, ConId, IName, FunName}),
+                             Impls
+                     end,
+    match_impls(Decls, ConId, IName, UnmatchedImpls).
+
+
+-spec compare_types(T, T) -> boolean() when T :: utype() | [utype()].
+compare_types(Types1 = [_ | _], Types2 = [_ | _]) ->
+    length(Types1) == length(Types2) andalso
+    lists:all(fun({T1, T2}) -> compare_types(T1, T2) end, lists:zip(Types1, Types2));
```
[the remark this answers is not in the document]
ghallak: I don't think that `compare_types` is needed, that was meant to be a version of `unify` that does not give a type error (only returns true or false).

If I use `unify` here instead of `compare_types`, I will be getting 2 errors, one `cannot_unify` error and another `unimplemented_interface_function` where I only need the second one.

I can modify `unify` to not give a type error in one of three ways:
1. Pass a value to `When` and use that to indicate that we don't need to throw an error in this case (Probably a bad idea since the value `When` is only used for output)
2. Add a flag to `Env` (e.g. `unify_no_thorw`) and only throw if it's set to `true`
3. Add an argument to `unify` (e.g. `ShouldThrow`) and only throw if it's set to `true` (That would cause changes in multiple lines of the code in each definition of `unify`)

test/aeso_compiler_tests.erl
ghallak: In the files `polymorphism_variance_switching.aes` and `polymorphism_variance_switching_custom_types.aes` there are tests for:

* custom datatypes
* tuples
* maps
* lists

I'll add to that tests for:
* options
* oracles
* oracle queries
* records
* type aliases

src/aeso_ast_infer_types.erl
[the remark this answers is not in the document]
ghallak: Actually, I have faced the same issue while working on a different PR. I think that adding something like `unify_throws = true` to the record `env` could be the best idea here.

src/aeso_ast_infer_types.erl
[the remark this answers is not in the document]
radrow: My point is to make it behave nicely when you use `I1` as an interface for a remote contract. Note that interfaces in Sophia are not abstract like in Java. 

Alternatively, the following code should work:

```
contract interface I0 =
  entrypoint f() : int

contract interface I1 : I0 =
  entrypoint something_else() : int

main contract C =
  entrypoint f(x : I1) = x.f() // Here we should know that x has f
```

src/aeso_ast_infer_types.erl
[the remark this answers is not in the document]
radrow: I think the previous name was better, this `_recursive` is mentioned in the comment anyway

src/aeso_ast_infer_types.erl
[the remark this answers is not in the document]
radrow: Are you sure that named args are sorted at this point? If not, you will have false-negatives on this check.

src/aeso_ast_infer_types.erl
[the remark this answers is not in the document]
radrow: What's wrong with unification errors here?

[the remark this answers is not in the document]
ghallak: I just want to check whether two types are the same or not, and move on if they are not. Unification errors will stop the compilation, and I don't want that.

src/aeso_ast_infer_types.erl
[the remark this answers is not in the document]
ghallak: I have renamed it to `check_implemented_interfaces1` here 1e965681cb595dac95828b0f3d2680f89e96f5a1

src/aeso_ast_infer_types.erl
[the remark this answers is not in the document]
ghallak: This is fixed now, check commits e74c75558b7a1fc5e48740f7f2f17ad7bbe2c76d and 831903e675091883688b43c1753aede892e4d8ab

ghallak:
I should probably rename all new test files to have a common prefix related to this PR before merging it.

src/aeso_ast_infer_types.erl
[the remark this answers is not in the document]
ghallak: I have implemented that here 28bf14cc7cbda129dc7d4525e642b145698c1fea

test/aeso_compiler_tests.erl
ghallak: I have tests for all the above now after 114010b8e1ff9a978912f736bd143364666bd89c, 8e995429aa6bd184b3d690f97dacf3ea6fb1b555, eab8eae9c1b97174fa93cb31201805aa45a4a578, and a6f1d8107713baa0379ecbacd06c1e51b4bd112c.

test/aeso_compiler_tests.erl
ghallak: Done that here da6d22717cf3410b97792894aba371f353078ea1

test/contracts/polymorphism_variance_switching.aes
```diff
@@ -0,0 +1,66 @@
...
+    stateful function f11(x : list(Cat)) : list(Animal) = g11(x)
+
+    stateful function g12(x : Animal * Animal) : Cat * Cat = (Chain.create(), Chain.create())
+    stateful function f12(x : Cat * Cat) : Animal * Animal = g12(x)
+
+    stateful function g13() : map(Cat, Cat) = { [Chain.create()] = Chain.create() }
+    stateful function f13() : map(Animal, Animal) = g13()
+
+    stateful function g2(x : Cat) : Cat = Chain.create()
+    stateful function f2(x : Animal) : Animal = g2(x)
```
radrow, line 39: I guess this is supposed to fail? Maybe make a note

test/contracts/polymorphism_variance_switching.aes
```diff
@@ -0,0 +1,66 @@
...
+    stateful function f3(x : Cat) : Cat = g3(x)
+
+    stateful function g4(x : (Cat => Animal)) : Cat = Chain.create()
+    stateful function f4(x : (Animal => Cat)) : Animal = g4(x)
+
+    stateful function g44(x : list(list(Cat) => list(Animal))) : Cat = Chain.create()
+    stateful function f44(x : list(list(Animal) => list(Cat))) : Animal = g44(x)
+
+    stateful function g5(x : (Animal => Animal)) : Cat = Chain.create()
+    stateful function f5(x : (Cat => Cat)) : Animal = g5(x)
```
radrow, line 51: this fails as well, right?

test/contracts/polymorphism_variance_switching_custom_types.aes
```diff
@@ -0,0 +1,132 @@
+contract interface Animal =
+    entrypoint sound : () => string
+
+contract Cat : Animal =
+    entrypoint sound() = "meow"
+
+main contract Main =
+    datatype ta('a) = TA('a => unit)
```
radrow, line 8: Would you make the names more descriptive? Don't have to be elaborate, but so it is not necessary to constantly lookup their defs

test/contracts/polymorphism_variance_switching_records.aes
```diff
@@ -0,0 +1,50 @@
+contract interface Animal =
+    entrypoint sound : () => string
+
+contract Cat : Animal =
+    entrypoint sound() = "meow"
+
+main contract Main =
+    record rec_a('a) = { x : 'a }
```
radrow: make names descriptive

test/contracts/polymorphism_variance_switching_records.aes
```diff
@@ -0,0 +1,50 @@
...
+    record rec_b('a) = { x : 'a => unit }
+    record rec_c('a) = { x : () => 'a}
+    record rec_d('a) = { x : 'a => unit,
```
radrow: missing bivariant

test/contracts/polymorphism_variance_switching_oracles.aes
```diff
@@ -0,0 +1,47 @@
...
+        let o09 : oracle(Cat, Animal)    = oracle() : oracle(Animal, Animal)  // success
+        let o10 : oracle(Cat, Animal)    = oracle() : oracle(Animal, Cat)     // success
+        let o11 : oracle(Cat, Animal)    = oracle() : oracle(Cat, Animal)     // success
+        let o12 : oracle(Cat, Animal)    = oracle() : oracle(Cat, Cat)        // success
+        let o13 : oracle(Cat, Cat)       = oracle() : oracle(Animal, Animal)  // fail
+        let o14 : oracle(Cat, Cat)       = oracle() : oracle(Animal, Cat)     // success
+        let o15 : oracle(Cat, Cat)       = oracle() : oracle(Cat, Animal)     // fail
+        let o16 : oracle(Cat, Cat)       = oracle() : oracle(Cat, Cat)        // success
+
+        let q01 : oracle_query(Animal, Animal) = query() : oracle_query(Animal, Animal)  // success
```
radrow, line 30: Imo oracle queries should be covariant on both positions:

- the latter type carries a result that can be retrieved by `Oracle.get_answer`
- the former type carries the question as a plain value under `Oracle.get_question`

src/aeso_ast_infer_types.erl
radrow: ...and now I am happy that we don't have recursive datatypes

test/contracts/polymorphism_variance_switching_oracles.aes
ghallak: Done here c767cd48a4c6399de89cd0eda9046fcd82e92f25

test/contracts/polymorphism_variance_switching_custom_types.aes
```diff
@@ -0,0 +1,132 @@
...
+    datatype dt_co('a) = DT_CO(unit => 'a)
+    datatype dt_inv('a) = DT_INV('a => 'a)
+    datatype dt_biv('a) = DT_BIV(unit => unit)
+    datatype dt_inv_sep('a) = DT_INV_SEP_A('a => unit) | DT_INV_SEP_B(unit => 'a)
+    datatype dt_co_nest_a('a) = DT_CO_NEST_A(dt_contra('a) => unit)
+    datatype dt_contra_nest_a('a) = DT_CONTRA_NEST_A(dt_co('a) => unit)
+    datatype dt_contra_nest_b('a) = DT_CONTRA_NEST_B(unit => dt_contra('a))
+    datatype dt_co_nest_b('a) = DT_CO_NEST_B(unit => dt_co('a))
+    datatype dt_co_twice('a) = DT_CO_TWICE('a => unit => 'a)
+    datatype dt_a_co_b_contra('a, 'b) = DT_A_CO_B_CONTRA('a => 'b => unit)
```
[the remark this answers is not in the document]
radrow: In Haskell both 'a and 'b would be contra if you write it this way

test/contracts/polymorphism_variance_switching_custom_types.aes
```diff
@@ -0,0 +1,132 @@
...
+        let vh3 : dt_contra_nest_b(Cat)    = DT_CONTRA_NEST_B(f_u_to_dt_contra_a)  // success
+        let vh4 : dt_contra_nest_b(Cat)    = DT_CONTRA_NEST_B(f_u_to_dt_contra_c)  // success
+
+        let vi1 : dt_co_nest_b(Animal) = DT_CO_NEST_B(f_u_to_dt_co_a)  // success
+        let vi2 : dt_co_nest_b(Animal) = DT_CO_NEST_B(f_u_to_dt_co_c)  // success
+        let vi3 : dt_co_nest_b(Cat)    = DT_CO_NEST_B(f_u_to_dt_co_a)  // fail
+        let vi4 : dt_co_nest_b(Cat)    = DT_CO_NEST_B(f_u_to_dt_co_c)  // success
+
+        let vj1 : dt_co_twice(Animal) = DT_CO_TWICE(f_a_to_u_to_a)  // success
+        let vj2 : dt_co_twice(Animal) = DT_CO_TWICE(f_a_to_u_to_c)  // fail
```
radrow, line 107: should pass

test/contracts/polymorphism_variance_switching_custom_types.aes
```diff
@@ -0,0 +1,132 @@
...
+        let vj3 : dt_co_twice(Animal) = DT_CO_TWICE(f_c_to_u_to_a)  // fail
```
radrow, line 108: should pass

test/contracts/polymorphism_variance_switching_custom_types.aes
```diff
@@ -0,0 +1,132 @@
...
+        let vj4 : dt_co_twice(Animal) = DT_CO_TWICE(f_c_to_u_to_c)  // success
+        let vj5 : dt_co_twice(Cat)    = DT_CO_TWICE(f_a_to_u_to_a)  // fail
+        let vj6 : dt_co_twice(Cat)    = DT_CO_TWICE(f_a_to_u_to_c)  // fail
```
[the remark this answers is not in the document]
radrow: But it should fail, just curious what's going on

test/contracts/polymorphism_variance_switching_records.aes
```diff
@@ -0,0 +1,51 @@
...
+        let r12   : rec_inv(Cat)    = rxcyc  // success
+
+        let rba : rec_biv(Animal) = { x = 1 }
+        let rbc : rec_biv(Cat)    = { x = 1 }
+        let r13 : rec_biv(Animal) = rba  // success
+        let r14 : rec_biv(Animal) = rbc  // success
+        let r15 : rec_biv(Cat)    = rba  // success
+        let r16 : rec_biv(Cat)    = rbc  // success
+
+        ()
```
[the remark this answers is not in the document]
radrow: The trick is that the last 3 rvalues must not be inferred to be `Cat`. It should be either `Animal` or a type error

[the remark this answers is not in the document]
radrow: If it's a type error different that "cannot unify Cat and Animal" add another round that should result in it:

```
let _ : xy(Cat) = {x = c, y = c} : xy(Animal) // success
let _ : xy(Cat) = {x = c, y = a} : xy(Animal) // fail
let _ : xy(Cat) = {x = a, y = c} : xy(Animal) // fail
let _ : xy(Cat) = {x = a, y = a} : xy(Animal) // fail
```

radrow (approved):

radrow:
It's not perfect, but it's usable. We should consider improvements in the inference, but what is important now is:

- System is correct (no barking cats)
- Whenever type inference fails, an explicit type annotation should solve the problem

hanssv (approved):
Looks ok to me! Please squash before merging 😅

PR #361, opened 2021-12-10
Clarify documentation on protected calls

hanssv:


velzevur (approved):

docs/sophia_features.md
```diff
@@ -99,6 +99,9 @@ running out of gas it is necessary to set a gas limit using the `gas` argument.
 However, note that errors that would normally consume all the gas in the
 transaction still only uses up the gas spent running the contract.
 
+Any side effects (state change, token transfers, etc.) made by a failing
+protected call is rolled back, just like they would be in the unproteced case.
```
uwiger, line 103:
```suggestion
protected call is rolled back, just like they would be in the unprotected case.
```

Issue #354, opened 2021-11-11
More flexible error formatting


hanssv:
When errors (and warnings?) are put together they are prepared for presentation as a string. This is not always great, for example:

https://github.com/aeternity/aesophia_http/issues/80
and https://github.com/aeternity/aesophia_http/issues/79

We should move the formatting further down the line (`pp` function in `aeso_errors`?) - and perhaps provide more ways to present the raw error-structures.

ghallak:
@hanssv Moving the formatting down the line (in this case moving `mk_error` from `aeso_ast_infer_types` to `aeso_errors`) would allow throwing errors in their original form (e.g. `{unbound_variables, {id, _, "x"}}` instead of `#err` record) and it would then be possible for `aesophia_http` to write its own version of `mk_error` to make errors formatting suitable for the output of `aesophia_http`.

Another way would be to write 2 versions of the errors messages that is produced by `mk_error`, the current one, and one that is suitable for the output of `aesophia_http` (without new lines and without line and col numbers) and have a command line option to choose which type of formatting do we want.

The first method would require making changes to `aesophia_http` when new types errors are added, and the second might bloat the code, especially if more types of formatting are added in the future.

PR #362, opened 2021-12-20
Simplify error messages reported by the compiler

ghallak:
Fixes: #354

This PR changes most error messages to:
1. Remove formatting newlines (between sentences).
2. Remove trailing newlines (they will only show when print the error message using `aeso_errors:pp` but not in `#err.msg`).
3. Remove position info (e.g. `at line x, col y`) when it's already available from `#err.pos`.
4. Surround code with \` \` to make it distinguishable from the rest of the error message ("Unbound variable `x`", instead of "Unbound variable x").

### DIFF
--- a/src/aeso_ast_infer_types.erl
+++ b/src/aeso_ast_infer_types.erl
@@ -2941,7 +2941,7 @@ mk_error({mismatched_decl_in_funblock, Name, Decl}) ->
     Msg = io_lib:format("Mismatch in the function block. Expected implementation/type declaration of ~s function", [Name]),
     mk_t_err(pos(Decl), Msg);
 mk_error({higher_kinded_typevar, T}) ->
-    Msg = io_lib:format("Type ~s is a higher kinded type variable "
+    Msg = io_lib:format("Type `~s` is a higher kinded type variable "
                         "(takes another type as an argument)", [pp(instantiate(T))]
                        ),
     mk_t_err(pos(T), Msg);
@@ -2954,7 +2954,7 @@ mk_error({unnamed_map_update_with_default, Upd}) ->
     Msg = "Invalid map update with default",
     mk_t_err(pos(Upd), Msg);
 mk_error({fundecl_must_have_funtype, _Ann, Id, Type}) ->
-    Msg = io_lib:format("~s was declared with an invalid type ~s. "
+    Msg = io_lib:format("`~s` was declared with an invalid type `~s`. "
                        "Entrypoints and functions must have functional types"
                        , [pp(Id), pp(instantiate(Type))]),
     mk_t_err(pos(Id), Msg);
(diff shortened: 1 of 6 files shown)


dincho:
IMO the compiler lib should not do any formatting at all but provide a list of structured errors downstream, that is to the HTTP and CLI compilers.

dincho:
As an example of very hard to parse error message is:
```
➜  code_errors git:(master) aesophia_cli polymorphic_query_type.aes
Code generation error in 'polymorphic_query_type.aes' at line 3, col 5:
Invalid oracle type
  oracle('a, 'b)
The query type must not be polymorphic (contain type variables).
```

A lot of newlines.

Another example duplicating position information and non-sense newlines:
```
➜  contracts git:(master) aesophia_cli 05_greeter.aes 
Type error in '05_greeter.aes' at line 48, col 10:
The contract Greeter (at line 48, column 10) has no entrypoints. Since Sophia version 3.2, public
contract functions must be declared with the 'entrypoint' keyword instead of
'function'.

Type error in '05_greeter.aes' at line 53, col 5:
Use 'entrypoint' for declaration of blockheight (at line 53, column 5):
  entrypoint blockheight : (unit) => uint
  ```

hanssv:
@dincho I think this is a good compromise, you can get a raw(ish) form of the error from the compilier, and you can ask the compiler to format it further. Or you can get the information in JSON. The reason for keeping it all in `aesophia` is to make updates/additions less cumbersome - otherwise you'd have to remember there is a new error when updating aesophia_cli/http, etc.

I don't see any major issues with the provided examples (they should no longer have duplicated position information!?) - separating different errors with a newline makes sense I think?

hanssv (approved):

ghallak:
@dincho Could you go over the changes to the files `test/aeso_compiler_tests.erl` and `src/aeso_ast_infer_types.erl`, to see if there are any error (or warning) messages that you think could be changed to be better?

dincho:
> separating different errors with a newline makes sense I think?

that's for sure, but I still see multiple newlines in a "single" error messages in the tests cc @ghallak 

that's my main issue, for a practical example I'm building a sublime plugin that parses the CLI output, however using \n for error messages separator (which is fine, and used and all the compilers I'm aware of), the usage of \n in the error message itself makes it un-parsable.

hanssv:
The primary purpose of the errors output by `aesophia_cli` is to be _read by a human_ rather than _parsed by a tool_! Perhaps we could add a switch (`--raw_errors` or something similar) to `aesophia_cli`?

And no, putting `\n` into a string doesn't make it unparseable but that discussion we've had before 😅

dincho:
> The primary purpose of the errors output by `aesophia_cli` is to be _read by a human_ rather than _parsed by a tool_! Perhaps we could add a switch (`--raw_errors` or something similar) to `aesophia_cli`?
> 
> And no, putting `\n` into a string doesn't make it unparseable but that discussion we've had before 😅

Humans are lazy and use tools .. 😅

dincho:
> And no, putting `\n` into a string doesn't make it unparseable but that discussion we've had before

Well, define that in regex, then we talk again :)

hanssv:
Well... `parsing != regex` 😈 
 
Normally you use `/s` or `?s` as modifier to do multiline regex?

hanssv:
Let's merge this, and I'll tackle @dincho's issue in a follow up PR (and `aesophia_cli`).

PR #364, opened 2022-01-05
Add oneline error pretty-printing

hanssv:
Some tools (and editors) prefer oneline error messages - so let's provide that in the simplest possible way.


ghallak (approved):
I think this should be enough to change all error messages into single line messages.

PR #367, opened 2022-02-02
Fix warnings reporting and stdlib warnings

ghallak:


priv/stdlib/Bitwise.aes
```diff
@@ -116,16 +116,6 @@ namespace Bitwise =
         1 => ubxor__(a / 2, b / 2, val * 2, acc + val)
         _ => ubxor__(a / 2, b / 2, val * 2, acc)
 
-  // Bitwise combined 'and' and 'not' of second argument for positive integers
-  // x 'bnand' y = x 'band' ('bnot' y)
-  // The tricky bit is that after negation the second argument has an infinite number of 1's
-  // use as many as needed!
-  //
-  // NOTE: this function is not symmetric!
-  private function ubnand(a, b) =
```
ghallak, line 125: @hanssv This function is private and not used at all in the `Bitwise` namespace, so it's showing a warning. Do we need it?

hanssv: No, it seems superfluous

hanssv (approved):

radrow (dismissed):

radrow (approved):

PR #368, opened 2022-02-04
docs: order namespaces alphabetically and place Set in includables

marc0olo:
I had to deal with `Set` recently and saw that it wasn't correctly placed in the `includable namespaces` section.

when fixing this I decided to sort everything alphabetically. makes sense IMO

### DIFF
--- a/docs/sophia_stdlib.md
+++ b/docs/sophia_stdlib.md
@@ -12,223 +12,39 @@ in the scope and do not need any actions to be used, while the others require so
 
 The out-of-the-box namespaces are:
 
+- [Address](#address)
+- [AENS](#aens)
+- [Auth](#auth)
 - [Bits](#bits)
 - [Bytes](#bytes)
-- [Char](#char)
-- [Int](#int)
-- [Map](#map)
-- [Address](#address)
-- [Crypto](#crypto)
-- [Auth](#auth)
-- [Oracle](#oracle)
-- [AENS](#aens)
-- [Contract](#contract)
 - [Call](#call)
 - [Chain](#chain)
+- [Char](#char)
+- [Contract](#contract)
+- [Crypto](#crypto)
+- [Int](#int)
+- [Map](#map)
+- [Oracle](#oracle)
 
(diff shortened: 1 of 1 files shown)


docs/sophia_stdlib.md
```diff
@@ -488,2003 +909,1586 @@ Oracle.check_query(o : oracle('a, 'b), q : oracle_query('a, 'b)) : bool
...
+#### from\_list
+```
+from_list(cs : list(char)) : string
+```
 
-If you expect getting calls with malformed `frac`s in your contract, you should use
-this function to verify the input.
+Converts a list of characters into a normalized UTF-8 string.
 
-## <a name='set-stdlib'>Set</a>
```
ghallak: I have added this `<a name='set-stdlib'>` because without it `#set` will direct you to `Bits.set` instead of `Set`

docs/sophia_stdlib.md
```diff
@@ -12,40 +12,233 @@ in the scope and do not need any actions to be used, while the others require so
...
-- [Pair](#pair)
-- [Triple](#triple)
 - [Bitwise](#bitwise)
 - [BLS12_381](bls12_381)
+- [Func](#func)
 - [Frac](#frac)
+- [List](#list)
+- [Option](#option)
+- [Pair](#pair)
 - [Set](#set-stdlib)
```
ghallak: I see that you have deleted `set-stdlib` name from below. Are you sure that clicking this will direct you to the right place?

ghallak (changes requested):

docs/sophia_stdlib.md
marc0olo: good point, I re-added this as we have other sub-headings in the `Bits` section where `set` is already present. I fixed it now and tested locally

docs/sophia_stdlib.md
marc0olo: yeah, should now work again with correct structure

ghallak (approved):

radrow (approved):

Issue #230, opened 2020-02-20
Reverse application operator |>


radrow:
This idea was wandering around for several months. Speaking of the `|>` operator like in Elixir, Elm and F#. This is super cool and useful:
```
function enterprise_pipeline(int x) : stuff = x
  |> IntegerDivider.divide_int().by(3)
  |> Int.to_str
  |> StringAnalyzer.analyze("casual")
  |> StringAnalyzerResultProcesserFactory.make_factory
  |> StringAnalyzerResultProcesserFactory.produce_stuff
```

PR #371, opened 2022-02-28
Introduce reverse application operator |>

ghallak:
Fixes: #230


ghallak:
Runtime tests: https://github.com/ghallak/aeternity/blob/3054792ba66e4086075af9713b064c8eb8d673f9/test/contracts/pipe_operator.aes

ghallak:
For consistency, should this be called pipe operator or reverse application operator? :thinking:

radrow (approved):
I think it would be nicer to make it behave just like in Elixir, so you could do
```
[1,2,3,4] |> List.flat_map(x => [x, -x])
```
Note the "partial" application of flatmap. Specifically, to let autocurry functions if needed (and fully applied). @ghallak @hanssv what do you think?

This could be a scope of another PR though

ghallak:
@radrow You suggestion seems good, but since we don't have partial application currently, this might be done in another PR.

radrow (approved):

hanssv (approved):

Issue #301, opened 2021-03-08
Call contract from contract with the same signature


cytadela8:
Currently one can call contract from another by using an variable with type of that contract like:

```
contract Remote =
  function get : () => int
  function can_resolve : (string, string) => bool

contract RemoteCall =

    function remote_resolve(r : Remote, name: string, key: string) : bool =
        r.can_resolve(name, key)
```

But what if we want to call another instance of a contract with the same code. Following example will result in an error:
```
contract Detector =
  entrypoint detect() : int=
    switch(Chain.block_hash(Chain.block_height))
      None => 1
      Some(_hash) => 2
  entrypoint call_detect(detector : Detector) : (int * int) =
    let d_1 = detector.detect()
    (detect(), d_1)

  entrypoint call_call_detect(detector1 : Detector, detector2 : Detector) : (int * int * int) =
    let (d_1, d_2) = detector1.call_detect(detector2)
    (detect(), d_1, d_2)
```

One has to use workaround like this:
```
@compiler >= 4.3

contract DetectorInterface2 =
  entrypoint detect : () => int

contract DetectorInterface =
  entrypoint detect : () => int
  entrypoint call_detect : (DetectorInterface2) => (int * int)

contract Detector =
  entrypoint detect() : int =
    switch(Chain.block_hash(Chain.block_height))
      None => 1
      Some(_hash) => 2
  entrypoint call_detect(detector : DetectorInterface) : (int * int) =
    let d_1 = detector.detect()
    (detect(), d_1)

  entrypoint call_call_detect(detector1 : DetectorInterface, detector2 : DetectorInterface2) : (int * int * int) =
    let (d_1, d_2) = detector1.call_detect(detector2)
    (detect(), d_1, d_2)
```

We should make it so the workaround is not needed.

ghallak:
@UlfNorell @hanssv Are there any reasons for why it is not possible to call a different instance of the same contract? Was this banned on purpose? Or should it be allowed?

hanssv:
I can't think of any reason and I think I remember us discussing adding it in the past - just never made it to the top of the list I guess 🤔 

I guess the hard question would be to figure out a good name for the implictly defined type then. `Self` or `This` might be popular options 😅

PR #379, opened 2022-04-28
Allow calling a different instance of the current contract

ghallak:
Fixes: #301, Fixes: #380

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - Options to enable/disable certain optimizations.
+- The ability to call a different instance of the current contract
+  ```
+  contract Main =
+    entrypoint spend(x : int) : int = x
+    entrypoint f(c : Main) : int = c.spend(10)
+  ```
 ### Changed
 ### Removed
 ### Fixed
--- a/src/aeso_ast_infer_types.erl
+++ b/src/aeso_ast_infer_types.erl
@@ -270,15 +270,24 @@ bind_state(Env) ->
         false  -> Env1
     end.
 
--spec bind_field(name(), field_info(), env()) -> env().
-bind_field(X, Info, Env = #env{ fields = Fields }) ->
+-spec bind_field_append(name(), field_info(), env()) -> env().
+bind_field_append(X, Info, Env = #env{ fields = Fields }) ->
     Fields1 = maps:update_with(X, fun(Infos) -> [Info | Infos] end, [Info], Fields),
     Env#env{ fields = Fields1 }.
 
--spec bind_fields([{name(), field_info()}], env()) -> env().
-bind_fields([], Env) -> Env;
-bind_fields([{Id, Info} | Rest], Env) ->
-    bind_fields(Rest, bind_field(Id, Info, Env)).
+-spec bind_field_update(name(), field_info(), env()) -> env().
+bind_field_update(X, Info, Env = #env{ fields = Fields }) ->
+    Fields1 = maps:update_with(X, fun([_ | Infos]) -> [Info | Infos]; ([]) -> [Info] end, [Info], Fields),
+    Env#env{ fields = Fields1 }.
+
+-spec bind_fields([{name(), field_info()}], typed | untyped, env()) -> env().
+bind_fields([], _Typing, Env) -> Env;
+bind_fields([{Id, Info} | Rest], Typing, Env) ->
+    NewEnv = case Typing of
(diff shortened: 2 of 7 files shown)


hanssv:
Looks promising - test failing?

ghallak:
The tests are failing because of this https://github.com/aeternity/aesophia/issues/380 which I'm working on fixing in this PR as well.

hanssv (approved):
Looks good! 

Maybet add an update to the CHANGELOG?

ghallak:
@hanssv I have updated the CHANGELOG here d84960274caca96e7b45c66870e3322c81aa6612

hanssv (approved):

CHANGELOG.md
```diff
@@ -11,6 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ```
   [1, 2, 3] |> List.first |> Option.is_some  // Option.is_some(List.first([1, 2, 3]))
   ```
+- The possibility to call different instance of the current contract
```
radrow, line 14: Ability? Possibility doesn't fit here imo

radrow, line 14:
```suggestion
- The ability possibility to call different instance of the current contract
```

src/aeso_ast_infer_types.erl
```diff
@@ -253,15 +253,22 @@ bind_state(Env) ->
         false  -> Env1
     end.
 
--spec bind_field(name(), field_info(), env()) -> env().
-bind_field(X, Info, Env = #env{ fields = Fields }) ->
+-spec bind_field(append | update, name(), field_info(), env()) -> env().
+bind_field(append, X, Info, Env = #env{ fields = Fields }) ->
```
radrow, line 257: these could be separate functions

src/aeso_ast_infer_types.erl
```diff
@@ -845,10 +861,11 @@ infer1(Env, [{Contract, Ann, ConName, Code} | Rest], Acc, Options)
         contract -> ets_insert(defined_contracts, {qname(ConName)});
         contract_interface -> ok
     end,
-    {Env1, Code1} = infer_contract_top(push_scope(contract, ConName, Env), What, Code, Options),
+    Env0 = bind_contract(untyped, Contract0, Env),
+    {Env1, Code1} = infer_contract_top(push_scope(contract, ConName, Env0), What, Code, Options),
     Contract1 = {Contract, Ann, ConName, Code1},
     Env2 = pop_scope(Env1),
-    Env3 = bind_contract(Contract1, Env2),
+    Env3 = bind_contract(typed, Contract1, Env2),
```
radrow, line 868: Do you really need to rebind it? I would expect the "untyped" version to unify to the final "typed" representation anyway. Do you check if they match?

CHANGELOG.md
ghallak: Done here d11e471bc2718356401a0444f5ae38ecaf3165be

src/aeso_ast_infer_types.erl
ghallak: If I don't rebind, the cases with type aliases will fail, since they are only replaced during the type inference.

src/aeso_ast_infer_types.erl
ghallak: Done here dba72071120f798a6174e64e286df0a0e828913c

src/aeso_ast_infer_types.erl
radrow: Could you give me some example? I also wonder if this could be hacked around somehow (that is, so both versions don't really match)

ghallak: This is one example:
```
contract interface Remote =
  record rstate = { i : int, s : string, m : map(int, int) }

  entrypoint fun_update_i  : (rstate, int) => rstate

contract StateHandling =

  type state = Remote.rstate

  entrypoint init(r : Remote, i : int) =
    let state0 = { i = 0, s = "undefined", m = {} }
    r.fun_update_i(state0, i)
```

The problem here is that the type of the arguments in the `fun_update_i` function declaration is set to `rstate` before the type inference and to `Remote.rstate` after the type inference (because the qualifications are added during type inference).

Binding the fields only before the type inference, would cause an error that the types `rstate` and `Remote.rstate` cannot be unified.

I can't currently think of a solution for this, so I may add a comment about it and maybe it can be handled later, unless you can think of a solution for this?

radrow: Maybe forcefully qualify all the types at every opportunity, so there is always `Remote.rstate`?

hanssv (approved):

radrow (approved):

radrow (approved):

hanssv (approved):

Issue #228, opened 2020-02-20
Add operator lambdas


radrow:
Ability to treat an operator in parentheses like a casual function, just like in Haskell and OCaml;
```
function sum(l : list(int)) : int = foldl((+), 0, l)
function plus(x, y) = (+)(x, y) // ugly but shows the idea
```

PR #385, opened 2022-06-01
Allow binary operators to be used as lambdas

ghallak:
Fixes: #228


test/contracts/operator_lambdas.aes
```diff
@@ -0,0 +1,14 @@
+include "List.aes"
+
+contract C =
+    type state = int
+
+    function sum(l : list(bool)) : bool = List.foldl((||), false, l)
```
radrow, line 6: sum? :p

src/aeso_parser.erl
```diff
@@ -472,6 +472,19 @@ id()       -> token(id).
 tvar()     -> token(tvar).
 str()      -> token(string).
 
+binop_as_lam() ->
+    BinOps = ['&&', '||',
```
radrow, line 476: It's sorta awkward that we maintain yet another lists of operators (one in lexer, one in expression parser and this one here)... Maybe there is a nicer way of doing so, so they won't desync?

Also, you don't want to allow `..` operator here, it doesn't make any sense to make it a lambda

radrow (approved):

test/contracts/operator_lambdas.aes
ghallak: Will change this to `any`.

src/aeso_ast_infer_types.erl
```diff
@@ -1641,7 +1641,10 @@ infer_expr(Env, {app, Ann, Fun, Args0} = App) ->
             GeneralResultType = fresh_uvar(Ann),
             ResultType = fresh_uvar(Ann),
             when_warning(warn_unused_functions,
-                         fun() -> register_function_call(Namespace ++ qname(CurrentFun), Name) end),
+                         fun() -> if element(1, Name) == lam -> ok;
+                                     true -> register_function_call(Namespace ++ qname(CurrentFun), Name)
```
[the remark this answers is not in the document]
ghallak: This was an existing bug that nobody has noticed. It didn't allow lambdas to be called directly on arguments (e.g. you can't do `((x) => x))(1)`).

I've noticed this when I was trying to do `(+)(1, 2)` and fixed it.

test/contracts/operator_lambdas.aes
ghallak: Changed it to `any` here 642f5a4da32090f2e1dd277b9d9b84482b3b8942

src/aeso_parser.erl
ghallak: Disabled `..` here 8851629fc7f0b1d8599c6f30166b0437857d570b

hanssv (approved):

Issue #229, opened 2020-02-20
Add comparable typevar constraints


radrow:
Currently all types can be combined with `==` and `=<`-like relation operators. While this works cool on ints, strings and possibly some other types, it seems to be completely useless for custom ADTs and functions. In my opinion the expression `((x, y) => x) == ((x, y) => y)` shouldn't typecheck, as it may lead to hard to find bugs. 
Another example: `FixedTTL(x) > FixedTTL(y)` works as expected comparing `x > y`, but if we accidentally compare `FixedTTL` and `RelativeTTL` the first one will be always greater, no matter of the arguments.

To solve this issue I propose adding qualification to typevars (similar to typeclasses, but restricted only to `Ord` and `Eq`):
```
// Most general type is spelled 'a. It cannot be compared and unified with generalized non-comparable typevar.
function id(x : 'a) : 'a = x

// Types that can be compared using `==` operator would have double tick before the tvar name
function neq(x : ''a, y : ''a) : bool = !(x == y)

// Types forming a linear order would require triple tick (or something else, this is starting to look ugly)
function lt(x : '''a, y : '''a) : bool = x < y
// ''a unifies with '''a
function spec_eq(x : '''a, y : '''a) : bool = x == y
```
The new syntax opens another way of dealing with it:
```
function 
  lt : 'a is ord, ('a, 'a) => bool
  lt(x, y) = x < y
```
I think this is the place where we could accept operator overloading. Although in situations where `==` can be derived automatically it is usually the best and the only solution, the `<` like operators can be more tricky so we could let the users define the comparison by themselves. Of course not stateful nor payable.

PR #390, opened 2022-06-14
Add comparable typevar constraints

ghallak:
Fixes: #229

### DIFF
--- a/src/aeso_syntax_utils.erl
+++ b/src/aeso_syntax_utils.erl
@@ -61,6 +61,7 @@ fold(Alg = #alg{zero = Zero, plus = Plus, scoped = Scoped}, Fun, K, X) ->
             {fun_t, _, Named, Args, Ret} -> Type([Named, Args, Ret]);
             {app_t, _, T, Ts}            -> Type([T | Ts]);
             {tuple_t, _, Ts}             -> Type(Ts);
+            {constrained_t, _, _, T}     -> Type(T);
             %% named_arg_t()
             {named_arg_t, _, _, T, E} -> Plus(Type(T), Expr(E));
             %% expr()


docs/sophia_features.md
```diff
@@ -501,6 +502,67 @@ function
...
+
+// Multiple constraints can be specified
+ord_and_eq : 'a is ord, 'b is eq ; ('a, 'a, 'b, 'b) => bool
+ord_and_eq(x, y, p, q) = x > y && p == q
+```
+
+If the constraints are not mentioned, the type variable is assumed to
+have no constraints, and calling any equality or inequality operation
+on it will cause an error.
+
```
radrow, line 565: I would use less "judging" language and more "factful". Also, restructured a little bit.

````suggestion
## Comparable types

Only certain types are allowed to be compared by equality (`==`, `!=`) and
inequality (`<`, `>`, `=<`, `>=`). For instance, while it is legal to compare
integers, comparing functions would lead to an error:

```
function f() = 
  f == f  // type error
```

The rules apply as follows:

- All types that are comparable by inequality are also comparable by equality.
- The builtin types `bool`, `int`, `char`, `bits`, `bytes`, `string`, `unit`,
  `hash`, `address` and `signature` are comparable by inequality (and thus by
  equality).
- The composite types `list`, `option`, and tuples are comparable by
  equality/inequality if their type parameters are comparable by
  equality/inequality.
- The composite types `map`, `oracle`, and `oracle_query` are comparable by
  equality if their type parameters are comparable by equality.
- User-defined records and datatypes are comparable by equality if their type
  parameters are comparable by equality.
- Smart contracts are comparable by equality.
- User-declared type variables are comparable according to the [type
  constraints](#type-constraints) given in the function signature.

In all other cases the types are not comparable.

### Type constraints

Polymorphic types are not declared as comparable by default. If the user
specifies the type signature for a function, they need to manually declare type
constraints in order to allow the variables to be compared. This can only be
done if the type declaration is separated from the function definition. The
constraints have to be prepended to the type declaration and separated with a
semicolon:

```

function eq(x : 'a, y : 'a) = x == y  // Type error, 'a is not comparable

function
  eq : 'a is eq ; ('a, 'a) => bool
  eq(x, y) = x == y                   // Compiles

function eq(x, y) = x == y            // Compiles as the constraints are inferred
```

Currently only two constraints are allowed: `eq` for equality and `ord` for
inequality. Declaring a type as `ord` automatically implies `eq`.

````

File: src/aeso_ast_infer_types.erl [hunk 2]
```diff
@@ -2329,6 +2368,33 @@ check_bytes_constraint(Env, {add_bytes, Ann, Fun, A0, B0, C0}) ->
...
+check_tvars_constraint(Env, {is_ord, Type = {uvar, Ann, _}}) ->
+    Type1 = unfold_types_in_type(Env, instantiate(Type)),
+    type_is_ord(Type1) orelse type_error({type_not_ord, Ann, Type1}).
+
+type_is_ord({app_t, _, Id, Ts}) -> type_is_ord(Id) andalso lists:all(fun type_is_ord/1, Ts);
+type_is_ord({tuple_t, _, Ts}) -> lists:all(fun type_is_ord/1, Ts);
+type_is_ord({bytes_t, _, _}) -> true;
+type_is_ord({constrained_t, _, Constraints, {tvar, _, _}}) -> lists:keyfind("ord", 3, Constraints) =/= false;
+type_is_ord({id, _, Id}) -> ets_lookup(ord_constraint_types, Id) =/= [];
+type_is_ord(_) -> false.
```
radrow, line 2388: After some thinking I think comparing addresses makes sense. It can be useful if you keep a list of addresses and want to ensure the order by sorting it

priv/stdlib/List.aes
```diff
@@ -29,9 +29,11 @@ namespace List =
     []   => abort("drop_last_unsafe: list empty")
 
 
-  function contains(e : 'a, l : list('a)) = switch(l)
```
radrow, line 32: Please add versions of `insert_by` and `sort` (and possibly others) that instead of taking a comparison lambda, rely on the constraints

radrow (commented):
Update fold in aeso_syntax_utils

priv/stdlib/List.aes
radrow: Remember that you actually can reuse the existing definitions by just applying them with a lambda

docs/sophia_features.md
```diff
@@ -357,7 +357,8 @@ Sophia has the following types:
 | Type                 | Description                                                                                 | Example                                                      |
 |----------------------|---------------------------------------------------------------------------------------------|--------------------------------------------------------------|
 | int                  | A 2-complement integer                                                                      | ```-1```                                                     |
-| address              | æternity address, 32 bytes                                                                 | ```Call.origin```                                            |
+| char                 | A single character                                                                          | ```'c'```                                                    |
+| address              | æternity address, 32 bytes                                                                  | ```Call.origin```                                            |
```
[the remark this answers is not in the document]
ghallak: Done here ea98fc97bbe2f8860850c1d812dd11b4b3ff8aa4

src/aeso_ast_infer_types.erl
```diff
@@ -2178,11 +2211,16 @@ destroy_and_report_unsolved_constraints(Env) ->
...
-    {BytesCs, []} =
+    {BytesCs, OtherCs4} =
         lists:partition(fun({is_bytes, _})              -> true;
                            ({add_bytes, _, _, _, _, _}) -> true;
                            (_)                          -> false
                         end, OtherCs3),
+    {TVarsCs, []} =
+        lists:partition(fun({is_eq, _})              -> true;
+                           ({is_ord, _}) -> true;
+                           (_)                          -> false
```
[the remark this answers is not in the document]
ghallak: Done here 9d296f04cb23d387fc0d6e3c47962ef166800591

src/aeso_ast_infer_types.erl
```diff
@@ -2016,7 +2048,8 @@ next_count() ->
 
 ets_tables() ->
     [options, type_vars, constraints, freshen_tvars, type_errors,
-     defined_contracts, warnings, function_calls, all_functions].
+     defined_contracts, warnings, function_calls, all_functions,
+     ord_constraint_types].
```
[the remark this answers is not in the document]
ghallak: There is no such table as I didn't find a use for it, if all `id`s are comparable by equality, what should go in that table?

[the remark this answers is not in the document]
radrow: I don't get why there is a need for a table for `ord` but not for `eq`. Would you explain why one is needed and the other is not?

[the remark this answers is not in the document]
ghallak: Since all types of the form `{id, _, Id}` are compared by `eq`, there is no need to store all of them in a table, but not all types of the above form are compared by `ord` so I only need to store that.

So checking if `{id, _, "int"}` is comparable by `ord` will only succeed because `"int"` is in the table `ord_constraint_types`, but checking if any other type of the form `{id, _, Id}` is comparable by `eq` will always succeed.

src/aeso_ast_infer_types.erl
ghallak: Done here fc2875965e302c75a1c02e9655beeb0c524aecee

ghallak: Will have to modify the tests for that.

src/aeso_ast_infer_types.erl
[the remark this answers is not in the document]
ghallak: Check this commit that made `address` comparable by `ord` fc2875965e302c75a1c02e9655beeb0c524aecee.

src/aeso_ast_infer_types.erl
ghallak: Fixed the tests here 4562a7166cfe7ddac65a940508a8266a0e7034f8

Issue #270, opened 2020-07-13
Wrongfully accepting interface definition in contract body


nikita-fuchs:
The compiler doesn't complain if you put a contract function interface somewhere into the contract, like https://studio.aepps.com?highlight=19-5-17-1&contract=036202a2-5 (corresponding tab doesn't focus automatically yet, you need to click it) 
When you try to deploy it, it of course throws you an error.

UlfNorell:
Please inline the code (or a shorter version) in the OP.

The contract only compiles if you call the compiler with the `no_code` flag. Otherwise
it gives you a nice error:
```
Code generation error in 'RemoteTest.aes' at line 18, col 16:
Missing definition of function 'lol_remote'.
```

nikita-fuchs:
I guess the `no_code` flag is being a default settings on the hosted compiler then, because the editor isn't calling any of that, 
```
    return this.http.post<any>(compilerUrl, {"code":`${code}`, "options":{}}, httpOptions);
```
summoning @dincho , who pointed me to this phenomenon 👋

radrow:
What does this flag do anyway?

UlfNorell:
It's supposed to be used when creating calldata and decoding return values and such, where you don't need a complete contract.

nikita-fuchs:
are we going to do anything about this ?

radrow:
@nikita-fuchs could you confirm if the issue is still present? Things have changed since the factories are in, maybe we can close this?

dincho:
> Generating ACI only does typechecking on the AST, and that is why you do not get the missing definitions error.

Should be this one

ghallak:
>could you confirm if the issue is still present?

@radrow I have checked in AE Studio, and the problem is still there. I just don't think that it's a problem with the compiler. I think we can close the issue after Nikita confirms that AE Studio is not using the `/compile` endpoint.

nikita-fuchs:
IIRC (and @hanssv might want to correct me just in case) we went for `/aci` because that is way faster than the full compilation process (which is fairly slow) and actually returns the `aci` that aestudio needs, which is not returned with `/compile` if I see correctly.

The good developer experience lives and dies with the fast responses we get from `/aci`. It's understandable that the crunching that takes place here is not that detailed like with `/compile`, but mistakes in contract code that result from bad syntax need to be shown while writing the contract, not only when it's being deployed. 
[...]

PR #394, opened 2022-06-24
Handle all user generated code errors in the type checker

ghallak:
Fixes: #270


src/aeso_ast_infer_types.erl
```diff
@@ -1612,8 +1629,51 @@ check_stateful_named_arg(#env{ stateful = false, current_function = Fun }, {id,
...
+is_monomorphic(Tup) when is_tuple(Tup) -> is_monomorphic(tuple_to_list(Tup));
+is_monomorphic(_)                      -> true.
+
+check_state_init(Env) ->
+    Top = Env#env.namespace,
+    StateType = lookup_type(Env, {id, [{origin, system}], "state"}),
+    case unfold_types_in_type(Env, StateType) of
+        false  ->
+            ok;
+        {_, {_, {_, {alias_t, {tuple_t, _, []}}}}} ->
```
radrow, line 1667: Why alias_t and not just tuple_t?

radrow (approved):

src/aeso_ast_infer_types.erl
ghallak: It's not possible to get this type with records and datatypes, so I added alias to make it clear that we're only matching `type state = ()`. I have added a comment here df9f258394823d74819df725a5c1ab6e6b5807eb to make that even clearer.

radrow (approved):

hanssv (approved):
Great stuff!

PR #396, opened 2022-06-26
Update the documentation and changelog to include polymorphism

ghallak:


docs/sophia_features.md
```diff
@@ -134,6 +134,63 @@ main contract IntHolderFactory =
 In case of a presence of child contracts (`IntHolder` in this case), the main
 contract must be pointed out with the `main` keyword as shown in the example.
 
+### Contract interfaces and polymorphism
+
+Contracts can implement one or multiple interfaces, the contract has to define
+every entrypoint from the implemented interface and the entrypoints in both
+the contract and implemented interface should have the exact same type.
```
radrow, line 141:
```suggestion
Contracts can implement one or more interfaces. For that, a contract has to implement
every entrypoint declared in that interface. The implementation has to match the type of the declaration in the interface.
```

...does it really need to be exact? Can't it be a supertype?

docs/sophia_features.md
```diff
@@ -134,6 +134,63 @@ main contract IntHolderFactory =
...
+
+ contract Cat : Animal =
+  entrypoint sound() = "Cat sound"
+```
+
+Contract interfaces can extend other interfaces, the extended interface has to
+declare all the entrypoints from the parent interface, and it can have
+additional entrypoints added to it. All the declarations in the extended
+interface must match (have the same type) with the declarations from the parent
+interface.
```
radrow, line 155:
```suggestion
Contract interfaces can extend other interfaces. An extended interface has to
declare all entrypoints from every parent interface. All the declarations in the extended
interface must have the same type as the declarations from the parent
interface.
```

docs/sophia_features.md
```diff
@@ -134,6 +134,63 @@ main contract IntHolderFactory =
...
+  entrypoint g : () => unit
+
+contract C : I =
+  entrypoint f() = ()
+  entrypoint g() = ()
+```
+
+It is only possible to implement (or extend) an interface that has been already
+defined earlier in the file (or in an included file), and therefore recursive
+interface implementation is not allowed in Sophia.
```
radrow, line 172:
```suggestion
defined earlier in the file (or in an included file). Therefore recursive
interface implementation is not allowed in Sophia.
```

docs/sophia_features.md
ghallak: It can be  a supertype, should I mention that?

radrow: Of course

docs/sophia_features.md
```diff
@@ -148,10 +148,9 @@ contract interface Animal =
   entrypoint sound() = "Cat sound"
 ```
 
-Contract interfaces can extend other interfaces, the extended interface has to
-declare all the entrypoints from the parent interface, and it can have
-additional entrypoints added to it. All the declarations in the extended
-interface must match (have the same type) with the declarations from the parent
+Contract interfaces can extend other interfaces. An extended interface has to
+declare all entrypoints from every parent interface. All the declarations in the extended
+interface must have the same type as the declarations from the parent
```
radrow, line 153:
```suggestion
interface must have types compatible with the declarations from the parent
```

radrow (dismissed):

docs/sophia_features.md
```diff
@@ -134,6 +134,94 @@ main contract IntHolderFactory =
...
+     entrypoint y : () => int
+     entrypoint z : () => int
+
+ contract C : Z =
+     entrypoint x() = 1
+     entrypoint y() = 1
+     entrypoint z() = 1
+```
+
+A function type `(Args1) => Ret1` is a subtype of `(Args2) => Ret2` when `Ret1`
```
radrow, line 194: Maybe format it as a list?

radrow: Something like

> The following subtyping rules apply:
>
> - blablabla
> - ...

docs/sophia_features.md
```diff
@@ -134,6 +134,94 @@ main contract IntHolderFactory =
...
+
+When a user-defined type `t('a)` is contravariant in `'a`, then `t(A)` is a
+subtype of `t(B)` when `B` is a subtype of `A`.
+
+When a user-defined type `t('a)` is binvariant in `'a`, then `t(A)` is a
+subtype of `t(B)` when either `A` is a subtype of `B` or when `B` is a subtype
+of `A`.
+
+When a user-defined type `t('a)` is invariant in `'a`, then `t(A)` can never be
+a subtype of `t(B)`.
```
radrow, line 224: Explain briefly variance below and link to wikipedia

radrow (approved):

docs/sophia_features.md
ghallak: Changed here 41c0609837588db08bb611cb2c19380216381e68

docs/sophia_features.md
ghallak: Done here 6c8715ea74d872471bf273816334cb14f627d19d

docs/sophia_features.md
ghallak: Done here 2ea56ce7efd2b49fa92b4520268c704103846f0f

docs/sophia_features.md
```diff
@@ -134,6 +134,104 @@ main contract IntHolderFactory =
...
+
+#### Subtyping
+
+The subtyping rules below, mention the variance of a type, which is defined in [Wikipedia](https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)) as follows:
+
+>Variance refers to how subtyping between more complex types relates to subtyping between their components.
+
+You can refer to the Wikipedia page for further understanding of type variance.
+
+In Sophia, the following subtyping rules apply:
```
radrow, line 202:
````suggestion
#### Subtyping and variance

Subtyping in Sophia follows common rules that take type variance into account. As described by [Wikipedia](https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)), 

>Variance refers to how subtyping between more complex types relates to subtyping between their components.

This concept plays an important role in complex types such as tuples, `datatype`s and functions. Depending on the context, it can apply to positions in the structure of a type, or type parameters of generic types. There are four kinds of variances:

- covariant
- contravariant
- invariant
- bivariant

A type is said to be on a "covariant" position when it describes output or a result of some computation. Analogously, position is "contravariant" when it is an input, or a parameter. Intuitively, when a part of the type is produced by values of it, it is covariant. When it is consumed, it is contravariant. When a type appears to be simultaneously input and output, it is described as invariant. If a type is neither of those (that is, it's unused) it's bivariant. Furthermore, whenever a complex type appears on a contravariant position, all its covariant components become contravariant and vice versa.

Variance influences how subtyping is applied. Types on covariant positions are subtyped normally, while contravariant the opposite way. Invariant types have to be exactly the same in order for subtyping to work. Bivariant types are always compatible.

A good example of where it matters can be pictured by subtyping of function types. Let us assume there is a contract interface `Animal` and two contracts that implement it: `Dog` and `Cat`.

```sophia
contract interface Animal =
  entrypoint age : () => int
  
contract Dog : Animal =
  entrypoint age() = // ...
  entrypoint woof() = "woof"
  
contract Cat : Animal =
  entrypoint age() = // ...
  entrypoint meow() = "meow"
```

The assumption of this exercise is that cats do not bark (because `Cat` does not define the `woof` entrypoint). If subtyping rules were applied naively, that is if we let `Dog => Dog` be a subtype of `Animal => Animal`, the following code would break:

```sophia
let f : (Dog) => string  = d => d.woof()
let g : (Animal) => string  = f
let c : Cat = Chain.create()
g(c)  // Cat barking!
```

That is because arguments of functions are contravariant, as opposed to return the type which is covariant. Because of that, the assignment of `f` to `g` is invalid - while `Dog` is a subtype of `Animal`, `Dog => string` is **not** a subtype of `Animal => string`. However, `Animal => string` **is** a subtype of `Dog => string`. More than that, `(Dog => Animal) => Dog` is a subtype of `(Animal => Dog) => Animal`.

This has consequences on how user-defined generic types work. A type variable gains its variance from its role in the type definition as shown in the example:

```sophia
datatype co('a) = Co('a) // co is covariant on 'a
datatype ct('a) = Ct('a => unit) // ct is contravariant on 'a
datatype in('a) = In('a => 'a) // in is invariant on 'a
datatype bi('a) = Bi // bi is bivariant on 'a
```

The following facts apply here:

- `co('a)` is a subtype of `co('b) when `'a` is a subtype of `'b`
- `ct('a)` is a subtype of `ct('b) when `'b` is a subtype of `'a`
- `in('a)` is a subtype of `in('b) when `'a` is equal to `'b`
- `bi('a)` is a subtype of `bi('b) always

That altogether induce the following rules of subtyping in Sophia:
````

radrow (approved):

hanssv (approved):

PR #398, opened 2022-07-02
Add loop operator in fcode

radrow:
Added `loop` construction in fcode to optimize list comprehensions and sequencing operator (and get rid of implicit imports)


radrow:
Good point Hans, recent progress in REPL allowed checking that after my """fix""" it uses 3x more gas XD Although I think it's a performance bug in my definition of `..`, not in the idea of loop itself

radrow:
Maybe it's because of the tuple?

PR #401, opened 2022-07-24
Prepare for new repl

radrow:
I had to expose some more functions and fix some typing issues


src/aeso_ast_infer_types.erl
```diff
@@ -155,6 +160,10 @@
 
 %% -- Environment manipulation -----------------------------------------------
 
+-spec switch_scope(qname(), env()) -> env().
```
ghallak, line 163: @radrow What do you need this function for?

radrow: This sets the env to work in a given scope (eg a namespace). `push_scope` wouldn't work, because it always starts with a clean scope.

REPL uses it for resolution of user-defined types. They have to be declared in  `?MOCK_CONTRACT`, and so while unfolding their definitions the typechecker has to be in the context of that contract (otherwise it complains that the type is unknown)

ghallak (dismissed):

hanssv (dismissed):
👍

ghallak (approved):

hanssv (approved):

Issue #271, opened 2020-07-29
Option to turn off the optimizer / certain optimizations


radrow:
Optimization is the place where the things may go wrong – ability to turn it off may nicely expose some internal bugs that appear from time to time. Also, this is a mandatory step for the debugger development.

PR #409, opened 2022-08-26
Add options to enable/disable certain optimizations

ghallak:
Fixes: #271 

The following options are introduced:

- `optimize_inliner`
- `optimize_inline_local_functions`
- `optimize_bind_subexpressions`
- `optimize_let_floating`
- `optimize_simplifier`
- `optimize_drop_unused_lets`
- `optimize_scode`

By default, all options are implicitly set to `true`, and they should be disabled separately by explicitly setting them to `false`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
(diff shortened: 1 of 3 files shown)


radrow:
I don't like optimize_all. I would just set all of them to true by default - I would avoid inconsistent information

hanssv (approved):

PR #411, opened 2022-09-01
Return mapping from variables to registers in fate compilation

ghallak:
This PR will change the result of the compilation to include a mapping from variables to fate registers.

Code example:

```
contract X =
  function g(z) =
    let g = 1
    g + z

  entrypoint f() =
    let g = g(5)
    let h = g + 2
    h + h

main contract C =
  entrypoint f(x) =
    let x = x + x
    let y = x + x
    let x = y + y
    x + x

  entrypoint g() =
    let x = 3
    x + x

  entrypoint q(x) =
    x + 1

  stateful entrypoint xz() =
    let aa : X = Chain.create()
    aa.f()
```

Registers in the compilation result:

```
variables_registers =>
 #{{"C","f","x"} => {var,2},
   {"C","f","y"} => {var,1},
   {"C","g","x"} => {var,0},
   {"C","q","x"} => {arg,0},
   {"C","x","a"} => {var,0},
   {"X",".X.g","g"} => {var,0},
   {"X",".X.g","z"} => {arg,0},
   {"X","f","g"} => {var,0},
   {"X","f","h"} => {var,1}},
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     entrypoint spend(x : int) : int = x
     entrypoint f(c : Main) : int = c.spend(10)
   ```
+- Return a mapping from variables to FATE registers in the compilation output.
 ### Changed
 ### Removed
 ### Fixed
--- a/docs/aeso_compiler.md
+++ b/docs/aeso_compiler.md
@@ -53,6 +53,8 @@ The **pp_** options all print to standard output the following:
 
 The option `include_child_contract_symbols` includes the symbols of child contracts functions in the generated fate code. It is turned off by default to avoid making contracts bigger on chain.
 
+The option `debug_info` includes information related to debugging in the compiler output. Currently this option only includes the mapping from variables to registers.
+
 #### Options to control which compiler optimizations should run:
 
 By default all optimizations are turned on, to disable an optimization, it should be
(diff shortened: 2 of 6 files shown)


radrow (dismissed):

ghallak:
@radrow Notice how the local function `X.g` is called `.X.g`. I'm using an existing function (`aeso_fcode_to_fate:make_function_name/1`) to get the functions names. I don't think we need this distinction in the names between local function and entrypoints. Should it just be called `g` instead of `.X.g`?

radrow:
Yes, let's make it just g. I think you can hack it around by calling that make name with the env set to entrypoint/nonprivate or something like that

ghallak:
I have made the change here f0db4fb96b9b6edcada96291d656a45c3eb1b924 and here is how the new output looks like:

```
variables_registers =>
 #{{"C","f","x"} => {var,2},
   {"C","f","y"} => {var,1},
   {"C","g","x"} => {var,0},
   {"C","q","x"} => {arg,0},
   {"C","x","a"} => {var,0},
   {"X","f","g"} => {var,0},
   {"X","f","h"} => {var,1},
   {"X","g","g"} => {var,0},
   {"X","g","z"} => {arg,0}}
```

radrow (dismissed):

radrow (dismissed):

hanssv (approved):
Looks good! Just checking, there is no usage of the compiler where the extra data is a burden? If so it might be an idea to make this configurable?

radrow (approved):

ghallak:
@hanssv I have added a compiler option that enables/disables the returning of the newly introduced map here ee37e187e892c32303370919ee048a21ec18e25f.

hanssv (approved):

Issue #413, opened 2022-09-03
Polymorphism is broken on init


radrow:
Repro:

```
payable contract interface SalesOffer =
    entrypoint init : (address, address, int, int) => void
    payable entrypoint do_a_backflip : () => bool

payable contract Test : SalesOffer =
  entrypoint init(_, _, _, _) = ()
  payable entrypoint do_a_backflip() = true
```

Error:

```
Unimplemented function `init` from the interface `SalesOffer` in the contract `Test`
```

Note that this should compile no matter of the return type of the init function (void is technically a subtype of everything).

PR #415, opened 2022-09-07
Polymorphism fixes

ghallak:
Fixes: #413

This introduces a couple of fixes for the referenced issue.

The commits should be merged without squashing.

### DIFF
--- a/src/aeso_ast_infer_types.erl
+++ b/src/aeso_ast_infer_types.erl
@@ -942,7 +942,7 @@ match_impls(_, [], _, _, Impls) ->
 match_impls(Env, [{fun_decl, _, {id, _, FunName}, FunType = {fun_t, _, _, ArgsTypes, RetDecl}} | Decls], ConId, IName, Impls) ->
     Match = fun({letfun, _, {id, _, FName}, Args, RetFun, _}) when FName == FunName ->
                     length(ArgsTypes) == length(Args) andalso
-                    unify(Env#env{unify_throws = false}, RetDecl, RetFun, unknown) andalso
+                    unify(Env#env{unify_throws = false}, RetFun, RetDecl, unknown) andalso
                     lists:all(fun({T1, {typed, _, _, T2}}) -> unify(Env#env{unify_throws = false}, T1, T2, unknown) end,
                               lists:zip(ArgsTypes, Args));
                ({fun_decl, _, {id, _, FName}, FunT}) when FName == FunName ->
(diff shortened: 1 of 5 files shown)


src/aeso_ast_infer_types.erl
```diff
@@ -909,55 +911,53 @@ infer1(Env, [{pragma, _, _} | Rest], Acc, Options) ->
...
+      Impls :: [aeso_syntax:con()],
+      DefinedContracts :: [aeso_syntax:decl()].
+populate_functions_to_implement(Env, ContractName, Impls, DefinedContracts) ->
+    create_type_errors(),
+    [ begin
+        Inserted = ets_insert_new(functions_to_implement, {name(Id), I, Decl}),
+        [{_, I2, _}] = ets_lookup(functions_to_implement, name(Id)),
+        Inserted orelse type_error({interface_implementation_conflict, ContractName, I, I2, Id})
+      end || {I, Decl = {fun_decl, _, Id, _}} <- functions_to_implement(Impls, DefinedContracts) ],
     destroy_and_report_type_errors(Env).
```
radrow, line 959: If that throws, is the newly created ets table cleared?

src/aeso_ast_infer_types.erl
```diff
@@ -1466,15 +1466,34 @@ check_reserved_entrypoints(Funs) ->
...
+    TypeSig = {type_sig, Ann, none, Named, Args, Ret},
+    delete_if_implementation(Env, Name, TypeSig),
+    {{Name, TypeSig}, {fun_decl, Ann, Id, Type1}};
 check_fundecl(Env, {fun_decl, Ann, Id = {id, _, Name}, Type}) ->
     type_error({fundecl_must_have_funtype, Ann, Id, Type}),
     {{Name, {type_sig, Ann, none, [], [], Type}}, check_type(Env, Type)}.
 
+%% Delete the function FunName with the signature FunSig from of functions
+%% to be implemented if it is included there, or return ok otherwise.
+-spec delete_if_implementation(env(), FunName, FunSig) -> true when
```
radrow, line 1478: I don't like this name. It is not clear from where you delete (the universe?). At first I thought you delete it from the contract. Why not something like "register implementation"?

src/aeso_ast_infer_types.erl
```diff
@@ -1466,15 +1466,34 @@ check_reserved_entrypoints(Funs) ->
...
+      FunName :: string(),
+      FunSig :: typesig().
+delete_if_implementation(Env, Name, Sig) ->
```
radrow, line 1481: Also, why do you match type sig as well?

src/aeso_ast_infer_types.erl
ghallak: It is cleared inside `clean_up_ets`. In which case would we need to clean the `functions_to_implement` table explicitly?

src/aeso_ast_infer_types.erl
ghallak: Changed here 2a55823633625f40d7143d48157e0844a976769c

src/aeso_ast_infer_types.erl
ghallak: Changed here 06879fdf7384fff7c038196a2f5799aefdfcd98a

radrow (approved):

hanssv (approved):

Issue #391, opened 2022-06-16
Hole expressions


radrow:
In Haskell and Scala there is an expression variant called "hole". It represents a yet-unimplemented expression that throws error on build. A nice feature of that is that it can be used to check what the typechecker expects in a given place:

```
entrypoint test(l) =
  List.head(List.map(???, l)) + 1 // note the question marks
```
```
Type error: found a hole of type 'a => int
```

I propose `???` as it doesn't clash with patterns

hanssv:
Just what every newly converted Solidity developer wants/needs, right? 🤣

PR #433, opened 2023-01-08
Add hole expression

ghallak:
Closes: #391


src/aeso_ast_infer_types.erl
```diff
@@ -2910,6 +2912,12 @@ unify1(_Env, {uvar, _, R}, {uvar, _, R}, _Variance, _When) ->
     true;
 unify1(_Env, {uvar, _, _}, {fun_t, _, _, var_args, _}, _Variance, When) ->
     type_error({unify_varargs, When});
+unify1(_Env, {id, Ann, "hole"}, B, _Variance, _When) ->
```
radrow, line 2915: This doesn't seem right. Hole is an expression, not a type. Holes should have normal types as anything else and an error should be reported in infer_expr.

radrow (changes requested):

radrow:
Doc pls

docs/sophia_features.md
```diff
@@ -560,6 +560,19 @@ Sophia has the following types:
...
+Hole expressions, written as `???`, are expressions that are used as a placeholder. During compilation, the compiler will generate a type error indication the type of the hole expression.
+
+```
+include "List.aes"
+contract C =
+    entrypoint f() =
+        List.sum(List.map(???, [1,2,3]))
+```
+
+The hole expression found on the last line of the above example will generate the error `` Found a hole of type `(int) => int` ``. Meaning that a function from `int` to `int` should be used instead of the `???` placeholder.
```
radrow, line 574:
```suggestion
A hole expression found in the example above will generate the error `` Found a hole of type `(int) => int` ``. This says that the compiler expects a function from `int` to `int` in place of the `???` placeholder.
```

radrow (approved):

radrow (approved):

PR #456, opened 2023-05-30
Add arbitrary size byte arrays

hanssv:
Introduce arbitrary sized binary arrays (type `bytes()`); adding `Bytes.split_any`, `Bytes.to_fixed_size`, `Bytes.to_any_size`, `Bytes.size`, `String.to_bytes`, and `Int.to_bytes`; and adjust `Bytes.concat` to allow both fixed and arbitrary sized byte arrays.
  
### DIFF
--- a/docs/sophia_stdlib.md
+++ b/docs/sophia_stdlib.md
@@ -371,7 +371,7 @@ Each bit is true if and only if it was 1 in `a` and 0 in `b`
 
 ### Bytes
 
-#### to_int
+#### to\_int
 ```
 Bytes.to_int(b : bytes(n)) : int
 ```
@@ -379,7 +379,7 @@ Bytes.to_int(b : bytes(n)) : int
 Interprets the byte array as a big endian integer
 
 
-#### to_str
+#### to\_str
 ```
 Bytes.to_str(b : bytes(n)) : string
 ```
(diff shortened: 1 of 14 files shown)


docs/sophia_stdlib.md
```diff
@@ -392,7 +392,8 @@ Returns the hexadecimal representation of the byte array
 Bytes.concat : (a : bytes(m), b : bytes(n)) => bytes(m + n)
 ```
 
-Concatenates two byte arrays
+Concatenates two byte arrays, if `m` and `n` are known at compile time, the
```
seanhinde, line 395: Would prefer:
""Concatenates two byte arrays. If `m` and `n` are..""

seanhinde (approved):

src/aeso_ast_infer_types.erl
```diff
@@ -3292,49 +3317,59 @@ create_freshen_tvars() ->
...
-freshen(Ann, [A | B]) ->
-    [freshen(Ann, A) | freshen(Ann, B)];
-freshen(_, X) ->
+freshen(Ann, {bytes_t, _, fixed}, Ctx) ->
+    X = fresh_uvar(Ann),
+    add_constraint({is_fixed_bytes, Ctx, X}),
+    X;
+freshen(Ann, {fun_t, FAnn, NamedArgs, Args, Result}, Ctx) when is_list(Args) ->
+    {fun_t, FAnn, freshen(Ann, NamedArgs, Ctx),
+     [ freshen(Ann, Arg, [{arg, Ix} | Ctx]) || {Arg, Ix} <- lists:zip(Args, lists:seq(1, length(Args))) ],
```
radrow, line 3346: You know there is an amazing function lists:enumerate which was added to stdlib by a well known and respected compiler developer?

radrow (approved):

src/aeso_ast_infer_types.erl
hanssv: Yes, but only in OTP-25 :-) we still support older versions I think!

PR #480, opened 2023-08-08
Improve constraint solving

hanssv:
Fingers itched... This solves one actual problem (on Ceres branch once rebased) where three constraints depended on each other and the current implementation only made two iterations.

The rewrite should be more general and easier to further change if necessary. Would be nice to separate solving and error reporting further - but lets save that for another rainy day.

### DIFF
--- a/src/aeso_ast_infer_types.erl
+++ b/src/aeso_ast_infer_types.erl
@@ -2609,22 +2609,19 @@ solve_constraints(Env) ->
                 end;
            (_) -> true
         end,
-    AmbiguousConstraints = lists:filter(IsAmbiguous, get_constraints()),
 
-    % The two passes on AmbiguousConstraints are needed
-    solve_ambiguous_constraints(Env, AmbiguousConstraints ++ AmbiguousConstraints).
+    solve_ambiguous_constraints(Env, lists:filter(IsAmbiguous, get_constraints())).
 
 -spec solve_ambiguous_constraints(env(), [constraint()]) -> ok.
 solve_ambiguous_constraints(Env, Constraints) ->
-    Unknown = solve_known_record_types(Env, Constraints),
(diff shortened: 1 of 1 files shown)


src/aeso_ast_infer_types.erl
```diff
@@ -2576,61 +2575,118 @@ get_constraints() ->
...
+
+    if length(UnsolvedCs) < length(Constraints) orelse Progress == true ->
+        solve_constraints_top(Env, UnsolvedCs);
        true ->
-            case solve_unknown_record_types(Env, Unknown) of
-                true -> %% Progress!
-                    solve_ambiguous_constraints(Env, Unknown);
-                _ -> ok %% No progress. Report errors later.
-            end
+        ok
```
radrow, line 2592: I guess this is when we did not improve anything. Why is it `ok` then?

radrow (commented):
It is hard for me to assess what exactly has changed since many functions have been moved around. Could you add some tests that show the improvement? 

Also, it's a great opportunity for adding a bit more comments around...

src/aeso_ast_infer_types.erl
hanssv: Feel free to rename it to `done` or something it doesn't really matter. This is the terminating case in the fix-point recursion.

hanssv:
@radrow The general problem addressed was the lack of a proper fix-point iteration - this shows up when there are constraints that depend on each other and the order in which you try suddenly matters. It seemed like you or @ghallak had already noticed this -- and "solved" it by adding a _second_ copy of all constraints to the solver. 

I could have solved the current issue by adding a third copy of the constraints, but instead opted to implement the iteration that checked for progress more generally. And yes there could have been more comments, but it could also have been done properly the first time.

UlfNorell (approved):

PR #483, opened 2023-08-23
Improve independence analysis in code optimizer

### LINKED ISSUE #482: Incorrect instruction order - optimization bug?
The following contract is incorrectly compiled:
```
contract C =
  record state = {m : map(int, int)}
  entrypoint init() = {m = {[1] = 2}}

  stateful entrypoint foo(ix : int) =
    let res = state.m[ix]
    __foo(ix)
    res

  stateful entrypoint __foo(ix : int) =
    put(state{ m = Map.delete(ix, state.m) })
```
The optimization phase will incorrectly push the `Map.lookup` (`let res = state.m[ix]`) past the call to `__foo` as is evident by the resulting FATE code:
```
...
FUNCTION foo( integer) : integer
  ;; BB : 0
          PUSH arg0
          CALL "IBB"
  ;; BB : 1
          POP var9999
          MAP_LOOKUP a store1 arg0
          RETURN
```

This results in any call to `foo` failing since either `ix` is not there in the first place, or we delete it before looking at it - this is obviously a bug.

hanssv:
There is an omission in the logic checking whether two byte code instructions are independent - as demonstrated by #482 

This adds a check that an instruction reading the store/state is not pushed past an impure instruction (for example a function call).

This results in:
```
FUNCTION foo( integer) : integer
  ;; BB : 0
          MAP_LOOKUP var0 store1 arg0
          PUSH arg0
          CALL "IBB"
  ;; BB : 1
          POP var9999
          RETURNR var0
```
Fixes #482 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 ### Fixed
 - Fixed a bug with polymorphism that allowed functions with the same name but different type to be considered as implementations for their corresponding interface function.
+- Fixed a bug in the byte code optimization that incorrectly reordered dependent instructions.
 
 ## [7.2.1]
 ### Fixed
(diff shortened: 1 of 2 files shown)


radrow (approved):
We definitely need runtime tests in the compiler repo.

ghallak (approved):

PR #489, opened 2023-09-07
Improve resolution of relative includes

### LINKED ISSUE #488: Sub-optimal include directory resolution
Not sure if it is a minimal reproducing example, but it should be small enough. Three files, in three different directories:
`/tmp/foo.aes`:
```
include "/tmp/dir1/bar.aes"
contract C =
  entrypoint f() = D.g()
```
`/tmp/dir1/bar.aes`:
```
include "../dir2/baz.aes"
namespace D =
  function g() = E.h()
```
`/tmp/dir2/baz.aes`:
```
namespace E =
  function h() = 42
```

Straight up compiling `foo.aes` fails:
```
Parse error in '/tmp/dir1/bar.aes' at line 1, col 1:
Couldn't find include file '../dir2/baz.aes'
```

The problem is that while resolving the includes in `bar.aes` (which is in `dir1`) we still do this relative to the root source file (i.e. in `/tmp`) this means we won't find `../dir2`.

A workaround is to pass both the top level and the subdirectory as include paths:
```
$ ./aesophia_cli /tmp/foo.aes -i /tmp -i /tmp/dir2
cb_+JdGA6DSagnFZzPzngQuBCq056yYPk6ObA8ElyhfEBqi/J/jPMC4arg9/gbU40wCNwAHAQNU/kTWRB8ANwA3ABoOgj8BAz/+eCFrswI3AAcEAxEG1ONM/uC+/GEANwAHBAMReCFrs6cvBBEG1ONMES5FLmgRRNZEHxFpbml0EXgha7MRLkQuZxHgvvxhBWaCLwCFNy40LjAA3jUBlw==
```

hanssv:
Fixes #488

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 ### Changed
-- Improve how includes with relative paths are resolved during parsing/compilation.
+- Improve how includes with relative paths are resolved during parsing/compilation. Relative
+  include paths are now always relative to the file containing the `include` statement.
 ### Removed
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 ### Changed
+- Improve how includes with relative paths are resolved during parsing/compilation.
 ### Removed
 ### Fixed
 
--- a/docs/sophia_features.md
+++ b/docs/sophia_features.md
@@ -493,6 +493,24 @@ the file, except that error messages will refer to the original source
 locations. The language will try to include each file at most one time automatically,
 so even cyclic includes should be working without any special tinkering.
 
+### Include files using relative paths
+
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 ### Changed
+- Improve how includes with relative paths are resolved during parsing/compilation.
```
radrow, line 10: I think it should be clearer on how they are improved. Also, this change should be documented - for example it is not clear which path takes priority if there is a file available relative to the current file and another one relative to cwd

radrow, line 10: Test needed

ghallak, line 10: Something like this explains the change in a better way.

```suggestion
- Resolve include paths relative to the including file path when the resolution relative to the root file path fails.
```

hanssv: Reformulated.

hanssv:
Added documentation + reformulated CHANGELOG.

Reworked the solution a bit, by always tracking the current src_dir in `Options`. This is slightly more verbose but a lot cleaner I think.

And added the missing test case from #488

radrow (approved):

ghallak (approved):

Issue #490, opened 2023-11-17
False positive "is defined but never used" warning if record depends on type


davidyuk:
I found this behaviour in aesophia_cli, but probably it is related to the whole compiler.

Reproduction.aes
```
main contract Test =
    type option_int = option(int)
    record option_point = {x: int, y: option_int}

    entrypoint test_option_record(a: option_point) = a
```
```
$ ./aesophia_cli --version                                                  
Sophia compiler version 7.4.0
$ ./aesophia_cli ./Reproduction.aes                       
Warning in './Reproduction.aes' at line 2, col 5:
The type `option_int` is defined but never used.
cb_+IZGA6BCUfwhvRBYY1PLAU4DgQk4WRXAe41d+BBRRcRYPGzQ1cC4WbD+RNZEHwA3ADcAGg6CPwEDP/7YGGWQADcBNwIHhwI3ADcBBzcCB4cCNwA3AQcBAQCkLwIRRNZEHxFpbml0EdgYZZBJdGVzdF9vcHRpb25fcmVjb3Jkgi8AhTcuNC4wAOFCQ8c=
```
The warning is not correct, because `option_int` is used to define `option_point`.

This appeared while working on https://github.com/aeternity/aepp-calldata-js/pull/251

Issue #491, opened 2023-11-19
False positive "is defined but never used" warning if event type in contract interface


davidyuk:
I found this behaviour in aesophia_cli, but probably it is related to the whole compiler.

Reproduction.aes
```
contract interface RemoteI =
  datatype event = RemoteEvent(int)
  entrypoint emitEvents : () => unit

contract Test =
  datatype event = Event(int)
  stateful entrypoint emitEvents(remote: RemoteI) : unit =
    Chain.event(Event(42))
    remote.emitEvents()
```
```
$ ./aesophia_cli --version                           
Sophia compiler version 7.4.0
$ ./aesophia_cli ./Reproduction.aes       
Warning in './Reproduction.aes' at line 2, col 3:
The type `event` is defined but never used.
cb_+OJGA6A8lEFdU+Y5bIcB/Y6CFyjQhDCni5at80+U1cCLiLmvDcC4tbiC/kTWRB8ANwA3ABoOgj8BAz/+ZaXgDwI3AYcBNwEHNwBGNgAAAGIvX58BgR9CGdutZoszd/R2GZyVS+3JJy1C+hSzitoW5feMmcSnAAEDP/7hKsRXADcBRwI3AAwDrwEAG1QCAxFlpeAPDwJvgibPDAMADAEAAwD8EeEqxFc3ADcAAK0vAxFE1kQfEWluaXQRZaXgDy1DaGFpbi5ldmVudBHhKsRXKWVtaXRFdmVudHOCLwCFNy40LjAAifgV3g==
```

I'm using this kind of contract to generate ACI. The event type provided in ACI is used to decode events emitted by `emitEvents`. I doubt that the compiler should warn about this.

ghallak:
@davidyuk Why do you think this is a bug?

The warning is shown because neither the datatype `event` nor the data constructor `RemoteEvent(int)` are used.

ghallak:
Closing this as it's the expected behavior and it's not a real bug.

davidyuk:
So, compiler suggest me to remove `event` definition in RemoteI, but I need it in ACI to parse the emitted event on the aepp side (actually, it is implemented in sdk). Seems that there is no way to rewrite contact to fix warning and keep aepp working, can you confirm that?

PR #492, opened 2023-11-27
Disable unused type warnings for type used inside of records

ghallak:
Fixes: #490, #491


ghallak:
@hanssv @radrow @davidyuk This PR fixes only #490 which I consider a real bug.

The issue #491 does not seem to be a bug for me, and it is the expected behavior from the compiler.

hanssv (approved):
👍

radrow (approved):

PR #501, opened 2024-02-01
Fix printing FCode with lifted lambdas

radrow:


ghallak (approved):
This seems to solve the issue with `pp_fcode` option.

hanssv (approved):
Looks good to me! Thanks!

PR #503, opened 2024-02-15
Allow self-qualification

hanssv:
Allow referring to `foo` in namespace `X` as `X.foo` when already inside `X`. 

This used to result in an error if dependencies happened to be analyzed in the wrong order...

This PR is supported by Æternity Foundation.

### DIFF
--- a/src/aeso_syntax_utils.erl
+++ b/src/aeso_syntax_utils.erl
@@ -119,11 +119,7 @@ used_ids(Ns, E) ->
     [ lists:last(Xs) || {{term, Xs}, _} <- used(E), in_ns(Xs, Ns) ].
 
 in_ns([_], _) -> true;
-in_ns(Xs, Ns) -> in_ns_(Xs, Ns).
-
-in_ns_([_], [])            -> true;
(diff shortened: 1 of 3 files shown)


src/aeso_syntax_utils.erl
```diff
@@ -110,8 +110,20 @@ fold(Alg = #alg{zero = Zero, plus = Plus, scoped = Scoped}, Fun, K, X) ->
...
 used_ids(E) ->
-    [ X || {{term, [X]}, _} <- used(E) ].
+    used_ids([], E).
+
+%% Used ids, top level or in (current) namespace
+used_ids(Ns, E) ->
+    [ lists:last(Xs) || {{term, Xs}, _} <- used(E), in_ns(Xs, Ns) ].
+
+in_ns([_], _) -> true;
+in_ns(Xs, Ns) -> in_ns_(Xs, Ns).
```
UlfNorell, line 122:
```erlang
in_ns(Xs, Ns) -> lists:droplast(Xs) == Ns.
```
?

UlfNorell (approved):

PR #507, opened 2024-04-29
Allow interfaces with defined state type

ghallak:
This PR fixes the issue that prevents the compilation of contract interfaces in which state type is defined. Example:

```
contract interface I =
  record state = { x : int }
  entrypoint f : () => state
  entrypoint init : () => state

main contract C =
  entrypoint g(c : I) : int = c.f().x
```


ghallak:
@hanssv Do you think that the following contract should compile? Should `init` only be required only when `clone` is called on the contract interface or should it be required for all contract interfaces?

```
contract interface I =
  record state = { x : int }
  entrypoint f : () => state

main contract C =
  entrypoint g(c : I) : int = c.f().x
```

hanssv:
> @hanssv Do you think that the following contract should compile? Should `init` only be required only when `clone` is called on the contract interface or should it be required for all contract interfaces where state is not `unit`?
> 
> ```
> contract interface I =
>   record state = { x : int }
>   entrypoint f : () => state
> 
> main contract C =
>   entrypoint g(c : I) : int = c.f().x
> ```

In general the interfaces only need to contain what is used in the main contract, so in that spirit I think it should compile, yes. But it wouldn't be the end of the world if it doesn't. But let's see if we can collect a few more opinions.

radrow:
Right now we have a special `void` type for cases where we don't really care what a thing is, but we want to make it unusable (like in the case of `init` in interfaces). I think we should stick to that and even more - enforce interface init to return void, since we have no way to check what is the type of the contract's actual state.

hanssv:
But for `clone` the type is needed, right? But since that is the only case where `init` is "useable" that could be the only place we require to list `init`?!

radrow:
For `clone` you only need arguments. The return type should remain opaque, however. I am thinking of making it either some explicitly "unknown" type such as `void` or `???`, or some hardcoded (and equally unusable) `NameOfTheInterface.state`

PR #509, opened 2024-05-24
Upgrade to the latest aebytecode to fix C warnings

ghallak:


radrow (approved):

hanssv (approved):
Would have preferred a separate PR with the version bump... But 🤷

PR #515, opened 2024-11-08
Update the docs

ghallak:
* Add `char` to the list of Sophia types

* Move `Chain.ttl` to stdlib docs instead of Sophia types

* Add a note about `Chain.ttl` being a special case

* Make `Call.gas_left` a constant instead of a zero-argument function

* Move the function `Int.mulmod` to its right place in stdlib docs

* Move the function `Crypto.poseidon` to its right place in stdlib docs

* Fix typos in 'sophia_stdlib.md'

* Add `bytes_t` to `utype()` in the type checker


hanssv (approved):
👍

radrow (approved):

PR #517, opened 2024-11-20
Docs: fix error in List.partition description

hanssv:
This PR is supported by Aeternity Foundation.

### DIFF
--- a/docs/sophia_stdlib.md
+++ b/docs/sophia_stdlib.md
@@ -1921,7 +1921,7 @@ Removes longest prefix from `l` in which all elements fulfill `p`.
 List.partition(p : 'a => bool, l : list('a)) : (list('a) * list('a))
 ```
 
-Separates elements of `l` that fulfill `p` and these that do not. Elements fulfilling predicate will be in the left/first list. For instance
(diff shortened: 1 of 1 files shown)


docs/sophia_stdlib.md
```diff
@@ -1921,7 +1921,7 @@ Removes longest prefix from `l` in which all elements fulfill `p`.
 List.partition(p : 'a => bool, l : list('a)) : (list('a) * list('a))
 ```
 
-Separates elements of `l` that fulfill `p` and these that do not. Elements fulfilling predicate will be in the right list. For instance
+Separates elements of `l` that fulfill `p` and these that do not. Elements fulfilling predicate will be in the left/first list. For instance
```
ghallak, line 1924:
```suggestion
Separates elements of `l` that fulfill `p` and these that do not. Elements fulfilling the predicate will be in the first element of the returned tuple. For instance
```

I think it's clearer this way.

hanssv: It is, but it isn't inline withthe rest of the documentation. `tuple` is rarely mentioned...

But let's use it anyway.

ghallak (approved):

radrow (approved):

PR #525, opened 2025-08-18
refactoring 

DigitaleZukunft:
refactoring


hanssv (changes requested):
Thanks for the PR, it seems to contain a lot of useful stuff, both w.r.t documentation and code structure. But there is also things that seems entirely unrelated like a script for counting lines of code, and adding `erlscripten` as a sub-directory.

I think this PR needs to be split into more sizeable chunks in order to be accesible for reviewers. Preferrably should each PR tackle one issue, and not result in 4k lines of codes changed 😅

PR #526, opened 2025-09-02
Type checker refactoring

ghallak:


src/typing/aeso_type_constraints.erl
```diff
@@ -0,0 +1,486 @@
...
+%%%-------------------------------------------------------------------
+
+-module(aeso_type_constraints).
+
+-export([ create_constraints/0
+        , add_constraint/1
+        , solve_all_constraints/1
+        , solve_then_destroy_and_report_unsolved_constraints/1
+        , freshen_type/3
+        , freshen_type_sig/3
```
ghallak, line 16: @radrow Do you think this should be here or somewhere else?

radrow: Those two should be moved. Freshening is an operation on types, has nothing to do with constraints.

src/typing/aeso_type_ets.erl
```diff
@@ -0,0 +1,103 @@
+-module(aeso_type_ets).
```
radrow, line 1: Those files need copyright I believe

radrow, line 1: Rename this file to `aeso_ets`, it has nothing to do with types

src/typing/aeso_type_ets.erl
```diff
@@ -0,0 +1,103 @@
...
+    tab2list/1,
+    tab2list_ordered/1,
+    tab_exists/1
+]).
+
+%% Named interface to ETS tables implemented without names.
+%% The interface functions behave as the standard ETS interface.
+
+init() ->
+    put(aeso_type_ets, #{}).
```
radrow, line 23: `aeso_type_ets` -> `?MODULE`

src/typing/aeso_type_ets.erl
```diff
@@ -0,0 +1,103 @@
...
+insert_ordered(Name, Object) ->
+    Count = next_count(),
+    TabId = tabid(Name),
+    ets:insert(TabId, {Count, Object}).
+
+tab2list_ordered(Name) ->
+    [E || {_, E} <- tab2list(Name)].
+
+next_count() ->
+    V = case get(counter) of
```
radrow, line 86: `counter` should be under some macro under a more opaque name

src/typing/aeso_type_ets.erl
```diff
@@ -0,0 +1,103 @@
...
+%% Cleanup
+
+ets_tables() ->
+    [options, type_vars, constraints, freshen_tvars, type_errors,
+     defined_contracts, warnings, function_calls, all_functions,
+     type_vars_variance, functions_to_implement].
+
+clean_up_ets() ->
+    [ catch delete(Tab) || Tab <- ets_tables() ],
+    ok.
```
radrow, line 103: Those should be moved to where they are initialized

src/typing/aeso_type_fmt.erl
```diff
@@ -0,0 +1,538 @@
+%%%-------------------------------------------------------------------
+%%% @copyright (C) 2018, Aeternity Anstalt
+%%% @doc
+%%%     Type error and warning formatting functions for Sophia type checker.
+%%%     This module provides error and warning message formatting functionality.
+%%% @end
+%%%-------------------------------------------------------------------
+
+-module(aeso_type_fmt).
```
radrow, line 9: rename to `aes_error_fmt` — this module formats errors and warnings, not types. Ideally, I would also add a module for constructing errors, just like we have in `aerepl`.

src/typing/aeso_type_helpers.erl
```diff
@@ -0,0 +1,260 @@
...
+        , type_error/1
+        , typesig_to_fun_t/1
+        , get_oracle_type/3
+        , pos/1
+        , pos/2
+        , create_options/1
+        , get_option/2
+        , when_option/2
+        , unqualify/2
+        ]).
```
radrow, line 41: I bet most of those can be moved to existing or better-named modules.

- `is_first_order`, `dereference`, `app_t`, `qid`, `name`, `fun_arity` and friends are queries and construction of types. Maybe just make a `aeso_types` module for those?
- `type_error` should go to an error-dedicated module
- `create_options` is an ETS thing — maybe it should be moved to where other ETS are initialised? Or maybe options deserve own module? Alternatively, one could argue they are part of the ENV

src/typing/aeso_type_warnings.erl
```diff
@@ -0,0 +1,197 @@
+%%%-------------------------------------------------------------------
+%%% @copyright (C) 2018, Aeternity Anstalt
+%%% @doc
+%%%     Type warning generation and tracking for Sophia type checker.
+%%%     This module provides warning creation functions and handles tracking
+%%%     of various unused constructs (includes, stateful, typedefs, etc).
+%%% @end
+%%%-------------------------------------------------------------------
+
+-module(aeso_type_warnings).
```
radrow, line 10: I'd combine this with errors.

radrow (approved):
Generally looks okay. Tests are passing. My comments regard minor improvements to the refactoring itself, I didn't find any dangerous looking changes.

hanssv (changes requested):
If 4k lines moved around was too big, then 4.5k lines moved around is also too big?!

I'm sure this is great stuff, but to review it in its current form would require time.

ghallak:
@hanssv This is a refactoring of a single file. The problem is that the file is 4400 lines of code, and that's what made the change this big 😅 

What I've done here is just moving related functionalities away from the main type checker module and into their own module.

Do you think it would be better if every split is done in its own PR? (splitting pretty printers, splitting warnings, etc.)

hanssv:
> Do you think it would be better if every split is done in its own PR? (splitting pretty printers, splitting warnings, etc.)

Yes, I think so, now in order to at least make a reasonable effort of a review I have 50 commits to look at (looking at the end result isn't feasible, it just says one big file removed and 4800 lines that may or may not come from the original file scattered around existing and new files). If that was split into 10 chunks of 5 commits each it would be easier for me 😅 

But I'll try to get there eventually, at the moment there are paying customers that need attention, so this is just not a priority right now.

hanssv (approved):
I wish I could have spent more time checking this, but as far as I can tell it looks fine!

Great job, sorry for the delayed response!

radrow:
My comments still apply. Merging so that it doesn't get outdated

